PR #6, opened 2022-10-20
BUG: Do not add another scale after min_length scale_factor reached

thewtex:
Results in a dask.array chunk size of 0, which causes errors.


thewtex:
Addresses: https://github.com/InsightSoftwareConsortium/itkwidgets/issues/554

PR #24, opened 2023-04-28
ENH: Add support for setting axis units from CLI

tbirdso:
commit c6a655b3d656c02a9a1ded784f1605b56f350f7a (HEAD -> support-units-input, tbirdso/support-units-input)
Author: Tom Birdsong <tom.birdsong@kitware.com>
Date:   Mon May 1 11:17:10 2023 -0400

    BUG: Fix rich input validation printouts to console

    Resolves issue where CLI would crash silently and not print expected
    validation messages to console when bad input parameters were parsed.

commit 6d48c12706cd22c69a8ce723626b6d1a94075942
Author: Tom Birdsong <tom.birdsong@kitware.com>
Date:   Fri Apr 28 17:11:38 2023 -0400

    ENH: Add support for setting axis units from CLI

    Allows setting units in output OME-Zarr file. Includes validation to
    ensure that units match existing allowable type hints.

    Addresses issue where output with "null" units could not be loaded in
    viewer applications such as Neuroglancer.

### DIFF
--- a/ngff_zarr/cli.py
+++ b/ngff_zarr/cli.py
@@ -16,7 +16,6 @@ from rich.panel import Panel
 from rich.live import Live
 from rich.pretty import Pretty
 from rich.spinner import Spinner
-from rich import print as rprint
 from zarr.storage import DirectoryStore
 import zarr
 import dask.utils
@@ -31,6 +30,7 @@ from .ngff_image_to_itk_image import ngff_image_to_itk_image
 from .detect_cli_io_backend import detect_cli_io_backend, ConversionBackend, conversion_backends_values
 from .methods import Methods, methods_values
 from .rich_dask_progress import NgffProgress, NgffProgressCallback
+from .zarr_metadata import is_dimension_supported, is_unit_supported
 from .config import config
 
 def _multiscales_to_ngff_zarr(live, args, output_store, rich_dask_progress, multiscales):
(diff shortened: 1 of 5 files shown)


tbirdso:
@thewtex Ping for review

ngff_zarr/cli.py
```diff
@@ -37,6 +37,7 @@ def main():
 
     metadata_group = parser.add_argument_group("metadata", "Specify output metadata")
     metadata_group.add_argument('-d', '--dims', nargs='+', help='Ordered OME-Zarr NGFF dimensions from {"t", "z", "y", "x", "c"}', metavar='DIM')
+    metadata_group.add_argument('-u', '--units', nargs='+', help='Ordered OME-Zarr NGFF axes spatial or temporal units', metavar='UNITS')
```
thewtex, line 40: To be more explicit on the comtmand line, can we have the same behavior as `--scale`, `--translation` here? i.e. `<dim1> <dim1value> <dim2> <dim2value> etc.`

thewtex (changes requested):
@tbirdso thank you!!

Please see comment inline.

CI is also failing on older Python. We may need a backport or alternative approach for `Iterator`.

ngff_zarr/cli.py
tbirdso: Good catch, updated.

tbirdso:
Hi @thewtex , I am surprised to see two baseline comparison tests failing in the most recent CI run. These changes are primarily intended to add units support to `cli.py` and should not generally impact multiscales conversions.

It looks like CI checks on `main` last ran a month ago, I am wondering if the issue lies in some dependency package update rather than this PR. Would you please re-run the `main` Python 3.9 check to verify whether this is the case? I do not have permissions to run pipelines on this repository.

https://github.com/thewtex/ngff-zarr/actions/runs/4538221765/jobs/7996917712

thewtex:
@tbirdso I did not see an option to restart on that previous passing run, but I added a commit that passes locally in #26

tbirdso:
> @tbirdso I did not see an option to restart on that previous passing run, but I added a commit that passes locally in #26

Thanks. Interesting... I'll dig in deeper here.

thewtex:
For what it is worth, CI is running well on `main` now.

tbirdso:
@thewtex Please try re-running CI

thewtex (approved):
@tbirdso excellent, thank you!

PR #30, opened 2023-05-20
BUG: Scale propagation fix

tbirdso:
Resolves an issue where improper input scalings were used in computing sigma values for dask image Gaussian blurring prior to downsampling. Under previous behavior input image was always considered to have input scale of 1.0 in each direction.


tbirdso:
Note that https://github.com/thewtex/ngff-zarr/issues/32 is blocking me from generating a new tarball with updated baseline results to reflect proper Gaussian blurring. If that issue is not observed on other systems then perhaps someone else could run `pytest` with these changes applied and update baselines appropriately.

thewtex:
@tbirdso baselines updated :+1:

PR #36, opened 2023-06-02
BUG: Set units dictionary from CLI instead of updating existing

tbirdso:
Resolves issue where attempting to resolve units input against a non-existent units dictionary from the input file would cause the CLI to error out.


thewtex (approved):
@tbirdso thank you!

PR #41, opened 2023-07-12
absolute vs. relative scales bug fix in _dim_scale_factors

### LINKED ISSUE #40: Multiscales bug
Seems like a scale level is being skipped? My expectation for this example is scale levels with sizes 512, 256, 128, and 64. But scale3 is 32. `ome_ngff_zarr.zarr/.zattrs` also shows voxel spacings of: 1, 2, 4, 16 - so consistent with the voxel grid sizes - but not what I thought I would get.

Screen Shot 2023-07-12 at 3 12 09 PM

GFleishman:
This fixes #40.

Commit history implies you were aware of this issue before, so make sure I didn't misunderstand anything:
https://github.com/thewtex/ngff-zarr/commit/7796b71db9831763c0bb066d6e2b6b79ce34e84b

I renamed `previous_dim_factors` to `previous_absolute_dim_factors`, since we need these numbers to be relative to the full resolution data, not relative to the previous scale level as before. Name change is good since `previous_absolute_dim_factors` is no longer equal to the previous iteration `dim_factors`.

This fixes my test case:
Screen Shot 2023-07-12 at 5 12 19 PM

### DIFF
--- a/ngff_zarr/methods/_dask_image.py
+++ b/ngff_zarr/methods/_dask_image.py
@@ -111,10 +111,12 @@ def _downsample_dask_image(
     ]
     previous_image = ngff_image
     dims = ngff_image.dims
-    previous_dim_factors = {d: 1 for d in dims}
+    previous_absolute_dim_factors = {d: 1 for d in dims}
+    previous_scale_factor = 1
     for scale_factor in scale_factors:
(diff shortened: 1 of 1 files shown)


thewtex (approved):
@GFleishman good find! LGTM. Thank you!

PR #47, opened 2023-07-19
More dimension_separator="/" added, so napari can open big data as well

GFleishman:
This fixes the second comment in: https://github.com/thewtex/ngff-zarr/issues/45


thewtex (approved):
@GFleishman thank you!! :1st_place_medal:

PR #71, opened 2024-03-04
ENH: Test macOS Apple Silicon (ARM) in CI

thewtex:


thewtex:
xref: https://github.com/actions/setup-python/issues/808

Issue #69, opened 2024-02-27
Add cucim OpenSlideIO API support to the CLI


thewtex:
This will possibly give us an all GPU pipeline from read to write with the cucim anti-aliasing and kvikio in #68.

CC @gigony

thewtex:
Also a flag to enable a `dask-cuda` cluster.

PR #72, opened 2024-03-05
Cucim reader

thewtex:


thewtex:
Addresses #69

thewtex:
We will take a different approach.

PR #85, opened 2024-08-22
Fix usages of Sets

btbest:
I tried running `to_multiscales` with `scale_factors=113` and experienced my script hanging.

While debugging I pinpointed that the `while` in `_ngff_image_scale_factors` is infinite looping.

This was because `while (sizes_array > double_chunks).any()` was comparing `[30 113 113]` to `[226 60 226]`. Or in some runs, `[226 226 60]` or `[60 226 226]`. In other words, `double_chunks` was randomly ordered.

I traced this back to the set `_spatial_dims = {"x", "y", "z"}` being used in comprehension expressions in a way that pulls values out of the set. Sets in Python are [not ordered](https://stackoverflow.com/questions/26420945/what-is-the-order-of-elements-in-a-python-set), and it looks like in practice I really was getting the values out of the set in a random order.

This PR fixes all usages of `_spatial_dims` that would create randomly ordered lists, but really this helper variable just shouldn't be a set if the rest of the code wants to have a helper in xyz order.

I added an additional early exit condition to the loop; this shouldn't really be necessary anymore with the other fixes, but it is a sane exit condition in every case.

I decided to contribute the fix and found that the corresponding test's asserts weren't in sync what the function currently produces 😅  I don't know what's intended, but I set the expected values to what it currently produces since it doesn't necessarily look wrong, at least for the already existing test cases. I added another test where the chunks aren't just regular polyhedra.

### DIFF
--- a/ngff_zarr/methods/_itk.py
+++ b/ngff_zarr/methods/_itk.py
@@ -223,7 +223,7 @@ def _downsample_itk_gaussian(
         previous_dim_factors = dim_factors
         previous_image = _align_chunks(previous_image, default_chunks, dim_factors)
 
-        shrink_factors = [dim_factors[sf] for sf in _spatial_dims if sf in dim_factors]
+        shrink_factors = [factor for dim, factor in dim_factors.items() if dim in _spatial_dims]
 
         # Compute metadata for region splitting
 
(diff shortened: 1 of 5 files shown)


thewtex (approved):
@btbest fantastic find!! :clap:

Thank you so much for the contribution!

I may remove ` or scale_factor == previous` in a follow-up.

PR #86, opened 2024-08-22
BUG: Add test for to_multiscales

btbest:
Related to my other PR, here is a test that illustrates some bugs in `to_multiscales` that I did not continue digging to fix...


btbest:
I'm just rolling by the example in the spec, where all transform vectors have one value per axis (and from a software perspective, since the vectors aren't axis-annotated, it's very murky if their lengths don't match the number of axes in `multiscales.axes`)

btbest:
Fixed with #88

Issue #103, opened 2024-11-20
to_multiscales() scale_factor argument question


kevinyamauchi:
Hello! I was trying to understand the `scale_factor` argument for `nz.to_multiscales()`. From the docs, I thought that if I give an single integer it would compute the scale factors that increase until the an edge in the image has a length smaller than the value. However, when I tried it, it didn't seem to do any downscaling (see example below). However, when I give scale factors as a list, it works as expected.

What is the expected behavior?

>scale_factors ([int](https://docs.python.org/3/library/functions.html#int) of minimum length, int per scale or [dict](https://docs.python.org/3/library/stdtypes.html#dict) of spatial dimension int's per scale) – If a single integer, scale factors in spatial dimensions will be increased by a factor of two until this minimum length is reached. If a list, integer scale factors to apply uniformly across all spatial dimensions or along individual spatial dimensions. Examples: 64 or [2, 4] or [{‘x’: 2, ‘y’: 4 }, {‘x’: 5, ‘y’: 10}]

To reproduce (ngff-zarr == 0.9.0):

```python
import ngff_zarr as nz
import numpy as np

image = np.zeros((50, 50, 50), dtype=float)
multiscales = nz.to_multiscales(image, scale_factors=4)

print(multiscales.scale_factors)

# returns an empty list, which I assume means no downscaling was applied.
# I would have expected the scales to be [2, 4]
[]
```

thewtex:
Hi @kevinyamauchi ,

Thank you for filing this issue!

Yes, the documentation can be improved to explain why we do not get any scaling for the case you outlined.

For:

```py
image = np.zeros((50, 50, 50), dtype=float)
multiscales = nz.to_multiscales(image, scale_factors=4, chunks=32)

print(multiscales.scale_factors)

# we get
[{'z': 2, 'y': 2, 'x': 2}]
```

This is because there is a constraint based on the chunk size.  Please take a look at #105

kevinyamauchi:
Got it. That makes sense! Thanks for the update to the docs!

thewtex:
@kevinyamauchi thanks again for reporting! 🙏

PR #105, opened 2024-11-21
DOC: Improve documentation of scale factors and interaction with checks

thewtex:
Re: #103


kevinyamauchi:
Thanks for the quick fix, @thewtex ! I think this clarifies how the scales are calculated

PR #126, opened 2024-12-23
channel

thewtex:
- ENH: tczyx, etc. support
- PERF: itkwasm downsampling in zyx order

### DIFF
--- a/.github/workflows/pixi-test.yml
+++ b/.github/workflows/pixi-test.yml
@@ -15,7 +15,7 @@ jobs:
 
       - uses: prefix-dev/setup-pixi@v0.8.1
         with:
-          pixi-version: v0.37.0
+          pixi-version: v0.39.3
           frozen: true
 
       - name: Run tests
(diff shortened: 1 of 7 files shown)


will-moore:
Thanks @thewtex,

I tried downsampling an image that has C dimension, hoping that the 2 separate channels would be downsampled independently, but it seems that the first channel is incorrectly resampled since it ends up being black:

```
import ngff_zarr as nz
import numpy as np

# 4D data
data = np.random.randint(0, 256, 16777216, dtype=np.uint8).reshape((2, 128, 256, 256))
image = nz.to_ngff_image(data, dims=['c', 'z', 'y', 'x'])

# downsample
multiscales = nz.to_multiscales(image, scale_factors=[2], chunks=64)

print("multiscales.images", multiscales.images)

# check that both channels have pixel data
for i, img in enumerate(multiscales.images):
    print("resolution", i)
    print("  ch0 max", img.data[0,:,:,:].max().compute())
    print("  ch1 max", img.data[1,:,:,:].max().compute())
```

This prints the following, showing that the first channel is black for the downsampled resolution:
```
resolution 0
  ch0 max 255
  ch1 max 255
resolution 1
  ch0 max 0
  ch1 max 211
```

thewtex:
@will-moore thanks for testing! I pushed a fix for that.

will-moore (approved):
Working for me now, thanks!

PR #136, opened 2025-02-19
Fixed the default translation assignment from 1.0 to 0.0.

odinsbane:
Included a test that fails before this fix, and passes afterwards.

### DIFF
--- a/test/test_to_multiscales_translate.py
+++ b/test/test_to_multiscales_translate.py
@@ -15,7 +15,7 @@ p, pixel location. t is the translate portion of the transform. s is the scale.
 Each scale will have it's own set of pixels, transform, and scale. pi, ti, si respectivly.
 s0 is the original scale factor, it is often used to describe the size of a voxel.
 
-The downsampling factor is the ration or original pixels to scaled pixels.
+The downsampling factor is the ratio of original pixels to scaled pixels.
 
     f = ( D0/D1, H0/H1, W0/W1 )
  
--- a/test/test_to_multiscales_translate.py
+++ b/test/test_to_multiscales_translate.py
@@ -3,7 +3,7 @@
 """
 # Multiscale transformation checks.
 
-This is a test to make sure that the transforms multiscales.
+This is a test to make sure that the transforms are propagated correctly to reduced scales in a multiscale representation.
 
 The coordnates represent the center of a pixel.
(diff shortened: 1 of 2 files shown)


thewtex:
@odinsbane :clap: 

CI found an issue

odinsbane:
Oops, I left an unused import in there. I also updated it so that there is a test_XXX method and pytest will run it. I ran the whole test suite too. It should be good now. :)

test/test_to_multiscales_translate.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python3
+
+"""
+# Multiscale transformation checks.
+
+This is a test to make sure that the transforms multiscales.
```
thewtex, line 6:
```suggestion
This is a test to make sure that the transforms are propagated correctly to reduced scales in a multiscale representation.
```

test/test_to_multiscales_translate.py
```diff
@@ -0,0 +1,110 @@
...
+
+The coordnates represent the center of a pixel.
```
thewtex, line 8:
```suggestion
The coordinates represent the center of a pixel.
```

test/test_to_multiscales_translate.py
```diff
@@ -0,0 +1,110 @@
...
+The transform will transform a coordinate to "real space".
+
+  r = s*p + t
+
+p, pixel location. t is the translate portion of the transform. s is the scale.
+
+Each scale will have it's own set of pixels, transform, and scale. pi, ti, si respectivly.
+s0 is the original scale factor, it is often used to describe the size of a voxel.
+
+The downsampling factor is the ration or original pixels to scaled pixels.
```
thewtex, line 18:
```suggestion
The downsampling factor is the ratio of original pixels to scaled pixels.
```

thewtex (changes requested):
@odinsbane excellent!

A few spelling notes inline.

odinsbane:
Thanks for the improvements. I wasn't sure if I should leave such a long description in. I also haven't used this feature of github, so if it ends up waiting on me you might have to say something. :D

thewtex:
@odinsbane thanks for the lovely test and descriptions! It may help more learners/contributors in the future.

thewtex (approved):
:100:

PR #215, opened 2025-09-01
Proof of concept validating OME-Zarr images with ome-zarr-models in tests

dstansby:
Because `ome-zarr-models` now supports OME-Zarr 0.5, and it is capable of performing full validation on top of what the JSON schemas are capable of validating, I thought it might be a nice idea to use `ome-zarr-models` to validate data written by `ngff-zarr`.

In this PR I've done a proof of concept validating one of the OME-Zarr 05 Image groups produced by `ngff-zarr`, and it passes 🎉 (which I'm considering equally a win for ome-zarr-models and ngff-zarr). If this is welcome, I'd be happy to roll this out to the other tests.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -32,6 +32,12 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -e "py/.[tensorstore]"
 
+      - name: Install ome-zarr-models
+        if:
+          ${{ matrix.python-version != '3.9' && matrix.python-version != '3.10'}}
+        run: |
+          python -m pip install -e "py/.[ome-zarr-models]"
+
       - name: Test with pytest
         run: |
           cd py && pytest --junitxml=junit/test-results.xml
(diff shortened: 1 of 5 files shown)


thewtex:
Congrats on the ome-zarr-models 1.0 release!

I think this test belongs in the ome-zarr-models repository instead of ngff-zarr.

dstansby:
Why do you think it belongs in the ome-zarr-models repository? My thinking was it's testing the creation routines of ngff-zarr, so it makes sense for it to live here.

thewtex:
ome-zarr-models can check whether its validation routines follow the community-defined spec.

dstansby:
ome-zarr-models is implemented so that it implements the specification, and can validate against the complete specification instead of just the JSON schemas, which cannot capture the full specification. For example, validating against the JSON schemas cannot check constraints such as "The "dimension_names" attribute MUST be included in the zarr.json of the Zarr array of a multiscale level and MUST match the names in the "axes" metadata."

So I thought it might be welcome here to use ome-zarr-models to run "full" validation against the specification on the output produced by ngff-zarr?

thewtex:
ngff-zarr implements the spec, not just the JSON schemas of the spec. We have a functional implementation that aims to follow the specification, not just the JSON schemas, and a rich test suite for validation. There may be errors or omissions, but it is fairly complete.  We aim to fulfill use cases and follow the specification as defined in the specification repository in the text and schemas and by the authors and contributors who helped create it.

> So I thought it might be welcome here to use ome-zarr-models to run "full" validation against the specification on the output produced by ngff-zarr?

No, ome-zarr-models is your interpretation / take of the specification.

PR #242, opened 2025-11-20
feat(hcs): add helpers and docs for writing .ozx

thewtex:
Documentation on how to generate a .ozx RFC-9 Zip OME-Zarr from a
plate, support for writing a small number of individual wells,
a new helper class, HCSPlateWriter, for writing with a context
manager in parallel, and better documentation and ergonomics
on converting a directory OME-Zarr store to a .ozx.

### DIFF
--- a/docs/hcs.md
+++ b/docs/hcs.md
@@ -199,6 +199,122 @@ After creating the plate structure, use `write_hcs_well_image` to write individu
 
 **Critical:** All wells in a plate must use the same OME-Zarr version (0.4 or 0.5). Mixing versions within a single plate is not supported.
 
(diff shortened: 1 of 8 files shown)


sebi06 (commented):
Hi @thewtex ,

that looks really nice :-) espcially for writing larger HCS plates.

What I was wondering if there is also a way to define the well layout inside the metadata. 

For example if one writes wells B2 & B3 - how do you know if that was a 6, 24, 48, 96, ... WellPlate?

I experimented a bit with such things here:

https://github.com/sebi06/omezarr_playground/blob/main/scripts_and_notebooks/ome_zarr_utils.py

Or did I overlook that in NGFF-ZARR?

Issue #98, opened 2024-10-22
Add RFC-5 Coordinate systems and transformations support


thewtex:
Re: https://ngff.openmicroscopy.org/rfc/5/index.html

xref: https://github.com/InsightSoftwareConsortium/itkwidgets/pull/757#issuecomment-2430107368

jo-mueller:
@thewtex thanks for the thoughts on this! I agree that recursive stuff can often lead to...unforeseen consequences. Having looked at #212 a bit more just now it has become a bit more clear to me how writing additional structures is intended to work. In the spirit of the newly added functions over there it is easier to 
- create the metadata structure as it is now (with the `coordinateSystems` and `coordinateTransformations` dataclasses)
- Write that top-level metadata structure to disk
- Add members to that spatial collection using a `to_transform_zarr` function that could directly check whether the images written to the collection share the coordinate systems that were previously defined at the top-level.

Here's a small MWE to write multiscales:

```python
import ngff_zarr as nz
import numpy as np
from ngff_zarr.v06.zarr_metadata import Scale, CoordinateSystem, Axis

x_scale = 2
y_scale = 2

coordinate_system_physical = CoordinateSystem(
    name="physical",
    axes=[
        Axis(name="y", unit="micrometer", type="space"),
        Axis(name="x", unit="micrometer", type="space"),
    ],
)

coordinateTransformation = Scale(
    scale=[y_scale, x_scale],
    input='',
    output=coordinate_system_physical,
)

ngff_image = nz.to_ngff_image(
    np.random.random((1000, 1000)),
    coordinateTransformation=coordinateTransformation,
    name='image1'
    )

ngff_multiscales = nz.to_multiscales(
    ngff_image,
    scale_factors=[2, 4])

nz.to_ngff_zarr('multiscale_example.ome.zarr', ngff_multiscales, version='0.6.dev1')
```

jo-mueller:
@thewtex small update/discussion point on this: The more I work on implementing RFC5/0.6 support in the ways of the current implementations, the more I find myself tangled up with deeply nested if/else checks and indexing into metadata of an unknown/diverse structure.

My conclusion at this point is that it's worth to put in some extra work in the metadata modelling to make sure that ngff-zarr can "absorb" larger, potentially breaking changes in the metadata in the future. This doesn't belong in the works for RFC5 implementation, hence I would propose the following roadmap:

[...]

PR #243, opened 2025-11-26
WIP: add basic support for RFC5 transformations

jo-mueller:
re #98 

Coded myself a bit into a corner over at #238 , so started from scratch with bringing RFC5-like metadata here. A bunch of tests are still failing, but I think for the time being some feedback on whether the path I pursued in implementing is how you (@thewtex ) would see it happening?

I think the biggest challenge is that right now, the two main supported versions (0.4 and 0.5) are quite similar in their metadata layouts so that a lot of the internal metadata parsing doesn't have to account for the actual version. I.e., `axes` is *always* under `multiscales > axes` and the only difference is the presence of a toplevel `ome` key. In 0.6, this will be quite different, and dragging the checks for the existence of `axes`, one or more `coordinateSystems` and `ome` through the whole code introduces a lot of boilerplate.

I don't have a good solution for this problem rn - the only thing I could think of would be a more elaborated metadata modelling with separate parsers for 0.4/0.5/0.6 metadata and, more importantly, conversion routines.

Anyway, would love to hear your thoughts on the matter 👍

### DIFF
--- a/py/ngff_zarr/v06/zarr_metadata.py
+++ b/py/ngff_zarr/v06/zarr_metadata.py
@@ -13,19 +13,19 @@ class CoordinateSystem:
 @dataclass
 class Scale:
     scale: List[float]
+    name: Optional[str]
+    input: Optional[Union[str, CoordinateSystem]]
+    output: Optional[Union[str, CoordinateSystem]]
     type: str = "scale"
-    name: Optional[str] = None
-    input: Optional[Union[None, str, CoordinateSystem]]
-    output: Optional[Union[None, str, CoordinateSystem]]
 
 
 @dataclass
 class Translation:
     translation: List[float]
+    name: Optional[str]
+    input: Optional[Union[str, CoordinateSystem]]
+    output: Optional[Union[str, CoordinateSystem]]
     type: str = "translation"
-    name: Optional[str] = None
-    input: Optional[Union[None, str, CoordinateSystem]]
-    output: Optional[Union[None, str, CoordinateSystem]]
 
 
 coordinateTransformations = Union[Scale, Translation]


py/ngff_zarr/v06/zarr_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+class TransformSequence:
+    input: Optional[Union[str, coordinateSystem]]
+    output: Optional[Union[str, coordinateSystem]]
+    transformations: List[coordinateTransformations]
+    type: str = 'sequence'
+    name: Optional[str] = None
+
+
+@dataclass
+class Dataset:
```
thewtex: This is quite a generic name -- what is this supposed to represent? Can we make it more specific?

py/ngff_zarr/from_ngff_zarr.py
```diff
@@ -117,12 +117,13 @@ def from_ngff_zarr(
...
-
-        dims = tuple(reversed(supported_dims))
+    if "axes" in metadata:
+        dims = [a["name"] if "name" in a else a for a in metadata["axes"]]
+    elif "coordinateSystems" in metadata:
+        dims = [a["name"] for a in metadata["coordinateSystems"][0]["axes"]]
     else:
-        dims = tuple(a["name"] if "name" in a else a for a in metadata["axes"])
+        from .v04.zarr_metadata import supported_dims
+        dims = list(reversed(supported_dims))
```
thewtex: We want dims to be a tuple (immutable)

thewtex (commented):
@jo-mueller beautiful, thank you!!

A few comments inline.

jo-mueller:
Need to rewrite after #250 is merged

PR #280, opened 2026-01-30
fix: handle non-power-of-2 scale_factors in to_multiscales

### LINKED ISSUE #264: `to_multiscale` fails with `scale_factors` not power of 2
I was experimenting with non-conventional scaling factor and I noticed that when the scaling factor is not a power of 2, the zarr array shapes are not correct. The ngff metadata is correct.

To reproduce:
```python
import ngff_zarr as nz
import numpy as np

# Generate pixel data (e.g. from a microscope, simulation, etc.)
data = np.random.randint(0, 256, size=(120, 120), dtype=np.uint8)
image = nz.to_ngff_image(data, scale={'y': 1.0, 'x': 1.0})

multiscales = nz.to_multiscales(image, scale_factors=[2, 3, 4])
nz.to_ngff_zarr('example.ome.zarr', multiscales)
image = nz.from_ngff_zarr('example.ome.zarr')
# original ok
assert image.images[0].data.shape == (120, 120)
# factor 2 ok
assert image.images[1].data.shape == (60, 60)
# factor 4 ok 
assert image.images[3].data.shape == (30, 30)
# factor 3 is wrong ==> should be (40, 40) but is (60, 60)
assert image.images[2].data.shape == (40, 40), image.images[2].data.shape
```
Version `ngff_zarr==0.20.1`

thewtex:
## Summary

Fixes #264: `to_multiscale` fails with `scale_factors` not power of 2.

When using non-power-of-2 scale factors like `[2, 3, 4]`, the resulting zarr array shapes were incorrect. For example, with a 120x120 input:
- Expected: `(120, 120)`, `(60, 60)`, `(40, 40)`, `(30, 30)`
- Actual: `(120, 120)`, `(60, 60)`, `(60, 60)`, `(60, 60)` ❌

## Root Cause

In `_prepare_next_scale()`, relative scale factors were computed using integer division:

```python
next_multiscales_factor = next_multiscales_factor // multiscales.scale_factors[index - 1]
```

For `[2, 3, 4]`, this produced:
- Scale 3: `3 // 2 = 1` → no downsampling
- Scale 4: `4 // 3 = 1` → no downsampling

## Solution

Modified `_prepare_next_scale()` to:

1. Use `_dim_scale_factors()` with original and previous image context to compute exact incremental factors
2. Check if incremental downsampling can achieve the target size
3. If incremental is possible, downsample from previous level (memory efficient)
4. If not possible, use the already-computed image from initial `to_multiscales()` call

Also fixed a pre-existing bug in `_compute_next_translation()` that caused `KeyError` for images with non-spatial dimensions (e.g., 'c' for channels).

## Changes

- `py/ngff_zarr/to_ngff_zarr.py`: Rewrite `_prepare_next_scale()` to correctly handle non-power-of-2 scale factors
- `py/ngff_zarr/methods/_dask_image.py`: Fix `_compute_next_translation()` to filter by spatial dimensions
- `py/test/test_non_power_of_2_scale_factors.py`: Add comprehensive regression tests

## Testing

Added tests covering:
- Basic non-power-of-2 scale factors (`[2, 3, 4]`)
- Parametrized tests with various combinations (`[3, 6, 9]`, `[2, 5, 10]`, etc.)
- 3D data with non-power-of-2 factors
- Per-dimension dict scale factors

All 205 existing tests continue to pass.

### DIFF
--- a/py/test/test_to_ngff_zarr_dask_image.py
+++ b/py/test/test_to_ngff_zarr_dask_image.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: MIT
 from ngff_zarr import Methods, to_multiscales, to_ngff_image, config
 import dask.array as da
-import pytest
 
 from ._data import verify_against_baseline
 
@@ -32,7 +31,7 @@ def test_gaussian_isotropic_scale_factors_two_components(input_images):
 
 def test_non_spatial_dims_with_auto_scale_factors():
     """Regression test for issue #328.
-    
+
     Tests that to_multiscales() handles non-spatial dimensions (e.g., 'c')
     correctly when using auto scale factors and the large image caching path.
     The issue occurred when dict-style scale factors introduced non-spatial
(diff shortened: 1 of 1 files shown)


jo-mueller:
Sounds like a good workaround to me :) Maybe `exact` would be a good name for the `fallback` option, too? Exact methods have a habbit of coming out a bit slower, so that may come across as intuitive. Will look at the code tomorrow, now it's late 😴

thewtex:
> How does the to_multiscales function manage to create an exact pyramid where the to_ngff_zarr doesn't? If exact downscaling isn't possible, shouldn't this be the case either way?

I am not sure what is meant here -- to_multiscales and to_ngff_zarr are working together and achieving the same outcome.

> Why does the pyramid level generation have to be exact? I.e., I think for most purposes the pyramid levels are a means for faster rendering and the integer flooring drift is something that's acceptable in this context?

Here we have to defer to the caller on what they want and why they want it. I agree that the test case, a small input, downscaling at 2,3,4, is not that useful for visualization, for example. What we can do is provide good defaults, e.g. scaling by a factor of 2 at each level. And most people will use the defaults :-).

PR #289, opened 2026-01-30
Fix dask 2026.1.2

melonora:
A few hours ago the new version of Dask got released that implements a change made by me with regards to `zarr_array_kwargs`. This used to be a separate argument as `.to_zarr` specified both array kwargs and read kwargs in `**kwargs`. In the latest version of dask that has changed so that it only includes array kwargs. This means that there was no need for a separate `zarr_array_kwargs` argument anymore. 

In anticipation of that PR I did not expose these arguments in the public API of ngff-zarr when I unpinned dask in ngff-zarr. This means that for the user here there won't be any change, other than that there is a dask constraint. The version must either be lower than 2025.12.0 or higher than 2026.1.2.

### DIFF
--- a/py/ngff_zarr/to_ngff_zarr.py
+++ b/py/ngff_zarr/to_ngff_zarr.py
@@ -609,32 +609,17 @@ def _write_array_direct(
         target = (
             zarr_array if (region is not None and zarr_array is not None) else store
         )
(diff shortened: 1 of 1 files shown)


thewtex:
Per chat, let's get this released to fix existing uses and follow-up with additional improvements.

PR #329, opened 2026-02-06
make robust against non-spatial dimensions

### LINKED ISSUE #328: `to_multiscales` fails for large data/default scale factors/Dask_IMAGE_GAUSSIAN
Consider this example dataset:

```python
import ngff_zarr as nz
import dask.array as da

dummy_array = da.random.random((4, 15000, 15000), chunks=(1, 1000, 1000))

ngff_image = nz.to_ngff_image(
    data=dummy_array,
    dims=("c", "y", "x"),
    scale={"c": 1.0, "y": 0.65, "x": 0.65},
    axes_units={"x": "micrometer", "y": "micrometer"},
)
```

This works:

```python
ngff_multiscales = nz.to_multiscales(
    data=ngff_image,
    scale_factors=[2, 4, 8],
    cache=False,  # Disable disk caching to avoid rechunking warnings,
    method=nz.Methods.DASK_IMAGE_GAUSSIAN
)
```

and this doesn't:

```python
ngff_multiscales = nz.to_multiscales(
    data=ngff_image,
    #scale_factors=[2, 4, 8],
    cache=False,  # Disable disk caching to avoid rechunking warnings,
    method=nz.Methods.DASK_IMAGE_GAUSSIAN
)
```

The problem doesn't occur for smaller array sizes (i.e., `4x512x512`) or smaller chunksizes or other downsampling methods.

jo-mueller:
Fixes #328 

Somehow the non-spatial dimensions (i.e. `c`) are lost along the way when the successive scaling factors are calculated. I'm kind of confused though why it only happens when the code takes the detour through `_large_image_serialization`, since the `dims` or `scale` values are not actually altered there 😕 

Anyway. this works 👍

### DIFF
--- a/py/ngff_zarr/methods/_dask_image.py
+++ b/py/ngff_zarr/methods/_dask_image.py
@@ -27,11 +27,11 @@ def _compute_next_scale(previous_image: NgffImage, dim_factors):
     return {
         dim: input_scale[dim] * dim_factors[dim]
         for dim in previous_image.dims
-        if dim in _spatial_dims
+        if dim in _spatial_dims and dim in dim_factors
     }
 
 
-def _compute_next_translation(previous_image, dim_factors):
+def _compute_next_translation(previous_image: NgffImage, dim_factors):
     """Helper method to manually compute output image physical offset.
         Note that this method does not account for an image direction matrix.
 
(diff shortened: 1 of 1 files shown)


thewtex (approved):
👍 

@jo-mueller thank you!!

PR #338, opened 2026-02-08
fix(ts): bump fizarrita for edge chunk handling

thewtex:
Zarr v3 edge chunks (at array boundaries) are stored at their actual
smaller size by spec-compliant writers (e.g., Python zarr). The codec
pipeline always reported shape=chunk_shape from metadata, causing
out-of-bounds reads and data corruption when decoding these smaller
chunks.

### DIFF
--- a/ts/deno.json
+++ b/ts/deno.json
@@ -34,7 +34,7 @@
     "@std/cli": "jsr:@std/cli@^1.0.27",
     "@std/http": "jsr:@std/http@^1.0.24",
     "@std/path": "jsr:@std/path@^1.1.4",
-    "@fideus-labs/fizarrita": "npm:@fideus-labs/fizarrita@^1.0.0",
+    "@fideus-labs/fizarrita": "npm:@fideus-labs/fizarrita@^1.0.2",
     "@fideus-labs/worker-pool": "npm:@fideus-labs/worker-pool@^1.0.0",
     "@zarrita/storage": "jsr:@zarrita/storage@^0.1.4",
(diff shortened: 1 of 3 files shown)


thewtex:
@mkitti This isn't related to the rectilinear chunk grid extension (zarr-developers/zarr-extensions#25). 

With the existing Zarr v3 regular chunk grid, chunks at array boundaries can be smaller than chunk_shape from metadata. For example, an array of shape [100] with chunk_shape: [64] has two chunks: chunk 0 is 64 elements and chunk 1 is 36 elements. Spec-compliant writers like Python zarr store that second chunk at its actual size (36 elements), not padded to 64.

The bug was that fizarrita's codec pipeline always reported shape = chunk_shape from metadata (i.e. 64) regardless of the actual decoded buffer size. This caused out-of-bounds reads and data corruption when decoding those smaller boundary chunks. The fix computes the actual edge chunk shape as min(chunk_shape[d], array_shape[d] - coord * chunk_shape[d]) per dimension and corrects the decoded chunk's shape/stride accordingly.

mkitti:
Maybe I will let @d-v-b but I think the spec compliant behavior with the regular chunk grid is for all chunks, including those extending beyond the array boundary, to be the same shape. If zarr-python is doing this in some cases, then that is not conforming to the specification.

Is there a difference between Zarr v2 and v3 here? I have been mainly thinking of the v3 case.

thewtex:
I think this only came up with Zarr v3.

PR #381, opened 2026-02-18
fix(ts): handle padded edge chunks and fix biased reservoir merge

thewtex:
Two fixes for chunk-size-dependent OMERO statistics errors:

1. omero_codec_worker: Add copySubRegion to fixEdgeChunkShapeStride so that
   padded edge chunks (written by setWorker with fill-value zeros beyond the
   valid region) have the valid sub-region extracted before statistics are
   computed.

2. mergeAccumulators: Fix biased reservoir sampling when merging accumulators
   from chunks of different sizes. Previously, each chunk contributed an equal
   number of samples (10,000) regardless of its actual voxel count, causing
   small edge chunks (with many background zeros) to be overrepresented.
   Now samples from each chunk proportionally to its population count using
   Fisher-Yates partial shuffle, producing unbiased quantile estimates.

### DIFF
--- a/ts/src/utils/compute_omero-shared.ts
+++ b/ts/src/utils/compute_omero-shared.ts
@@ -438,31 +438,49 @@ export function mergeAccumulators(
   a: ChannelStatisticsAccumulator,
   b: ChannelStatisticsAccumulator,
 ): ChannelStatisticsAccumulator {
-  const merged: ChannelStatisticsAccumulator = {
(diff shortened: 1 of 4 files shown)


thewtex:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/fideus-labs/ngff-zarr/pull/381#pullrequestreview-3817649977)

Issue #48, opened 2023-07-19
Dask task graphs are huge (non-performant) for large data


GFleishman:
I think this will be a tougher issue to solve than typos, math bugs, and formatting; so it will warrant some conversation with @thewtex and @tbirdso.

I have `ngff-zarr` running well on small in-memory sized datasets (e.g. 512x512x512 voxels). When formatted with the nested directory style dimension separator these files open just fine with the napari-ome-zarr plugin (just my first check for compatibility with other tools - I'll use other visualization software later as well). So my next task was to run on real data.

My real image is {'z': 1596, 'y': 8983, 'x': 5451} voxels and I am attempting the following:

```python
ngff_image = to_ngff_image(
    my_big_zarr_array,
    dims=('z', 'y', 'x'),
    scale={a:b for a, b in zip('zyx', my_voxel_spacing)},
    axes_units={a:'micrometer' for a in 'zyx'},
)

multiscales = to_multiscales(
    ngff_image,
    scale_factors=[{'z':4, 'y':8, 'x':8}],
    chunks=(64, 128, 128),
)

to_ngff_zarr(
    './my_data_as_ome_ngff_zarr.zarr',
    'multiscales,
)
```

Importantly, this does actually work. If I wait long enough I get the file I want. However there is a performance issue that IMO will be lethal for the package if not resolved. For this (4, 8, 8) downsample level, the task graph has about 2.5M tasks. The dask distributed scheduler really can't handle graphs of this size well, which it acknowledges with a warning that the task graph itself is about 220MB when compressed, which it knows is huge, and says it will cause long slow downs.

It takes about 45 minutes for the scheduler to finish parsing all those tasks (just about right with the 2.5M tasks and the [documented 0.001 seconds per task](https://distributed.dask.org/en/latest/efficiency.html)), and during this time the workers are allocated but idle. And this has to happen for every scale level. In a real scenario I might have 10 or more scale levels I want to create. So that's a huge amount of time to be paying for cpus that are just waiting for the scheduler to finish parsing the task graph.

I feel that there are three different options for approaching this problem:
(1) small hacks, maybe something will help
(2) optimize the task graph by reworking the series of dask array operations to simplify simplify simplify
(3) abandon dask in as many places as possible

I tried three different things for (1):
(a) ran `dask.array.optimize` on the big graph (only eliminated about 10K tasks and at the expense of running this function which was long)
[...]

PR #398, opened 2026-02-20
perf(py): add map_blocks fast path for bin-shrink downsampling

thewtex:
## Summary

Adds a `map_blocks` fast path for bin-shrink-based downsampling methods (`ITKWASM_BIN_SHRINK` and `ITK_BIN_SHRINK`) that produces a much smaller dask task graph when chunk sizes are favorable. This directly addresses the performance issue described in #48, where large images generate task graphs with millions of tasks that overwhelm the dask scheduler.

## Problem

When generating multiscale pyramids, the existing code path always calls `_align_chunks` (which may rechunk the entire array) and then iterates over every non-spatial dimension index individually, issuing a separate `map_blocks`/`map_overlap` call for each. For large images with many chunks, this produces enormous dask task graphs (2.5M+ tasks for a moderately sized volume), causing the dask distributed scheduler to spend 45+ minutes just parsing the graph while workers sit idle.

## Solution

When the shrink factor divides evenly into **every** chunk (including the last, possibly smaller chunk) along each spatial dimension, bin-shrink can operate on each chunk independently. In this case we skip `_align_chunks` and the per-non-spatial-dim iteration loop entirely, replacing them with a single `da.map_blocks` call. This produces a dramatically smaller task graph.

### New helper: `_can_use_map_blocks_fast_path()`

Added to `_support.py`. Checks whether all chunks along each spatial dimension are divisible by the corresponding shrink factor. Returns `False` if any chunk (including the last) would leave a remainder, causing a graceful fallback to the standard path.

### ITKWASM fast path (`_itkwasm.py`)

- New `_itkwasm_chunk_bin_shrink_nd()` wrapper that handles blocks with leading non-spatial dimensions (t, c) by iterating over `np.ndindex` and calling the spatial-only `_itkwasm_chunk_bin_shrink` on each sub-block.
- Fast-path branch inserted before the standard path in `_downsample_itkwasm()`. Checks three conditions: (1) smoothing is `bin_shrink`, (2) channel configuration is vector-compatible (≤8 channels or no trailing `c` dim), and (3) `_can_use_map_blocks_fast_path()` returns `True`.
- When all conditions are met, issues a single `map_blocks` call with pre-computed output chunk sizes, then rechunks to the requested `out_chunks`. Uses `continue` to skip the standard path entirely.

### ITK fast path (`_itk.py`)

- Same pattern but simpler, since `itk.bin_shrink_image_filter` handles spatial data natively.
- Additional guard: the fast path is skipped whenever there are **any** non-spatial dimensions (t, c), because `itk.image_view_from_array` only supports up to 4D and would fail at compute time with 5D blocks. The standard path (which raises `ValueError` for unsupported dimensionality) is preserved.

### Fallback behavior

When the fast-path conditions are not met (uneven last chunks, many-channel trailing `c`, non-spatial dims for ITK), the code falls through to the existing standard path with zero behavioral changes. The standard path is completely untouched.

## Tests

19 new tests in `py/test/test_bin_shrink_map_blocks_fast_path.py`:

- **7 unit tests** for `_can_use_map_blocks_fast_path`: even chunks, uneven last chunk, non-spatial dims ignored, factor-1 edge case, 3D cases, smaller-but-divisible last chunk
- **9 ITKWASM integration tests**: 2D, 3D, 5D (tczyx), vector mode (zyxc with ≤8 channels), many-channels fallback (>8 channels), dict-based scale factors, fast-path-vs-standard correctness comparison, custom output chunks, dask graph size comparison
- **3 ITK integration tests**: 2D, 3D, dask graph size comparison

Full test suite: **390 passed, 0 failed, 2 skipped** (zero regressions).

## Closes

Partially addresses #48 (reduces task graph size for the bin-shrink case; Gaussian smoothing methods remain unchanged).

### DIFF
--- a/py/ngff_zarr/methods/_itkwasm.py
+++ b/py/ngff_zarr/methods/_itkwasm.py
@@ -99,12 +99,16 @@ def _itkwasm_chunk_bin_shrink_nd(
     combination of non-spatial indices, apply bin-shrink to each spatial
     sub-block, and reassemble the result.
     """
-    if block.ndim == n_spatial_dims or (is_vector and block.ndim == n_spatial_dims + 1):
+    # n_spatial_dims counts only spatial axes; for vector images, there is an
+    # additional channel axis. Any remaining leading axes are treated as
+    # non-spatial and iterated over.
+    non_spatial_ndim = block.ndim - n_spatial_dims - (1 if is_vector else 0)
+    if non_spatial_ndim == 0:
+        # Block is exactly spatial (and optional channel for vector images),
+        # so we can call the underlying 2D/3D bin-shrink implementation
+        # directly.
         return _itkwasm_chunk_bin_shrink(block, shrink_factors, is_vector=is_vector)
 
-    non_spatial_ndim = block.ndim - n_spatial_dims
-    if is_vector:
-        non_spatial_ndim -= 1  # channel dim is part of vector image
     non_spatial_shape = block.shape[:non_spatial_ndim]
 
     # Determine output spatial shape from the first sub-block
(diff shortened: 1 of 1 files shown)


thewtex:
CC @jo-mueller

PR #403, opened 2026-02-23
fix(py): pass multithreading error in to_multiscales

DiloCorp:
add main thread condition to run signal code in to_multiscales._large_image_serialization


thewtex (approved):
@DiloCorp thank you for the contribution! 🥇

PR #480, opened 2026-04-15
fix(py): keep zarr_format for dask versions using top-level create_array

thewtex:
dask ~2026.3.0 changed from Group.create_array() (which inherits
zarr_format from the group) to top-level zarr.create_array() (which
defaults to format 3 if zarr_format is absent). This caused
"Expected a BytesBytesCodec" errors when writing format 2 arrays
with numcodecs compressors, since format 3 parsing rejects them.

Detect the dask code path at import time and only remove zarr_format
from kwargs when dask uses Group.create_array() (which rejects it).
When dask uses the top-level zarr.create_array(), keep zarr_format
so arrays are created with the correct format.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with Dask and Zarr v3 by enhancing runtime detection of Dask capabilities.
  * Fixed compression settings handling for Zarr v3 to ensure proper argument transformation across different Dask implementations.
  * Enhanced conditional handling of format parameters based on detected Dask feature support.

### DIFF
--- a/py/ngff_zarr/to_ngff_zarr.py
+++ b/py/ngff_zarr/to_ngff_zarr.py
@@ -40,6 +40,18 @@ zarr_version = Version(zarr.__version__)
 IS_ZARR_V3_PLUS = zarr_version.major >= 3
 DASK_SUPPORTS_SHARDING = Version(dask_version) >= Version("2025.12.0")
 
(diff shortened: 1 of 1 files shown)


thewtex:
CC @melonora

thewtex:
Superceded by #393

PR #523, opened 2026-06-01
refactor(py): cull all None metadata fields recursively

### LINKED ISSUE #496: Generalize removal of metadata optionals
Currently, in the `to_ngff_zarr` code, there is this function, which removes fields with value `None` from the metadata dictionary to be written, which makes a lot of sense:

```python
def _pop_metadata_optionals(metadata_dict, enabled_rfcs: list[int] | None = None):
    # Collect all axes that need cleaning
    axes_to_clean = []
    if "axes" in metadata_dict:
        axes_to_clean = metadata_dict["axes"]
    elif "coordinateSystems" in metadata_dict:
        axes_to_clean = [ax for cs in metadata_dict["coordinateSystems"] for ax in cs["axes"]]
    
    # Clean axes
    for ax in axes_to_clean:
        if not ax["unit"]:
            ax.pop("unit")
        if not is_rfc4_enabled(enabled_rfcs):
            ax.pop("orientation", None)

    # pop empty coordinateTransformations on top-level only if they are None
    if metadata_dict["coordinateTransformations"] is None:
        metadata_dict.pop("coordinateTransformations")

    if metadata_dict["omero"] is None:
        metadata_dict.pop("omero")

    return metadata_dict
```

I think it could make sense to replace this function by a small recursive function, that browses through nested dictionaries and removes any `None` field it encounters. That way, we can save ourselves the version-unagnostic data parsing that's happening there (and will happen more with future versions)#

Edit: The check for the enabled rfc should not matter in here anyway, should it? Because even if RFC4 is enabled, an rfc4-like field with the value `None` should still be removed, right?

thewtex:
## Summary

Closes #496.

`_pop_metadata_optionals` in `py/ngff_zarr/to_ngff_zarr.py` previously removed
`None` metadata fields by enumerating each optional field by hand (axis `unit`,
top-level `coordinateTransformations`, `omero`, plus RFC 4 `orientation`). That
approach has to be revisited every time a new OME-Zarr version or RFC adds a
field. This PR replaces it with a small recursive helper that culls **any**
`None`-valued key, making the cleanup version-agnostic.

As discussed in the issue, the OME-Zarr spec treats an absent field and an
explicit `null` as equivalent — only fields that are present carry meaning — so
stripping every `None` is safe.

## What changed

- **`_remove_none_values(obj)`** (new): recursively walks nested dicts/lists and
  drops every key whose value is `None`. Returns a new structure. Only `None` is
  removed — falsy-but-valid values (`0`, `0.0`, `""`, `False`, empty
  collections, and importantly zero `scale`/`translation` entries) are kept.
- **`_pop_metadata_optionals`**: reduced to (1) the RFC 4 orientation gate, then
  (2) a single `_remove_none_values(metadata_dict)` call. The per-field
  `pop(...)` logic is gone.

## Why the RFC 4 gate stays

The issue suggested the `enabled_rfcs` check might be unnecessary. It is still
required, but for a reason orthogonal to `None`-stripping: `orientation` is
populated onto axes back in `to_multiscales()`, independent of which RFCs are
enabled at write time. RFC 4 is a draft feature, so a **populated** orientation
must remain opt-in and is dropped unless `enabled_rfcs=[4]`. Removing the gate
would leak draft metadata for any image that happens to carry orientations
(`test_rfc4_integration_without_enabled_rfcs` asserts exactly this). The gate is
therefore kept; the recursive pass then handles the `None` case for both
enabled and disabled writes.

## Behavioral fix

With RFC 4 enabled on a 5D image, the non-spatial `t`/`c` axes used to serialize
`"orientation": null` (their orientation is `None`, and the old code only popped
orientation when RFC 4 was *disabled*). These keys are now culled instead of
written as `null`.

Standard output is otherwise unchanged: in practice `to_multiscales` always sets
a downsampling method, so top-level `type`/`metadata` are non-null, and the
existing baseline `DeepDiff` comparisons still match byte-for-byte.

## Tests

- **`py/test/test_metadata_none_removal.py`** (new): 7 unit tests for
  `_remove_none_values` (top-level / nested / in-list removal, falsy-value
  preservation, scalar passthrough) and `_pop_metadata_optionals` (RFC 4
  on/off).
- **`py/test/test_rfc4_integration.py`**: added
  `test_rfc4_no_null_orientation_on_non_spatial_axes` (regression for the fix
  above) and split `test_rfc4_integration_without_enabled_rfcs` into its own
  function — its `def` line was missing, so its gating assertions had been
  running as a non-isolated block inside the with-enabled test.

Verified locally (115 tests) across the new unit tests, all RFC 4 tests,
multiscales metadata/type, baseline-comparison tests
(`test_from_ngff_zarr`, `test_to_ngff_zarr_{dask_image,itkwasm,itk,tensorstore}`),
v0.5 / RFC 2, sharding, and RFC 9 `.ozx`. `ruff format` + `ruff check` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Metadata output now recursively removes null optional fields (including nested dicts/lists) while preserving falsy-but-valid values (0, "", False, empty lists/dicts). RFC4 anatomical orientation is handled correctly: orientation entries are removed when RFC4 is disabled, and non-spatial axes do not emit null orientation when RFC4 is enabled.

* **Tests**
  * Added unit and integration tests covering recursive metadata cleaning, nested structures, lists, falsy values, and RFC4 orientation behaviors.

### DIFF
--- a/py/ngff_zarr/to_ngff_zarr.py
+++ b/py/ngff_zarr/to_ngff_zarr.py
@@ -5,7 +5,7 @@ import tempfile
 import warnings
 from dataclasses import asdict
 from pathlib import Path, PurePosixPath
-from typing import Literal
+from typing import Any, Literal
 
 from .methods._metadata import get_method_metadata
 
@@ -106,22 +106,44 @@ def _numcodecs_to_zarr_v3_codec(compressor):
     return None
 
 
-def _pop_metadata_optionals(metadata_dict, enabled_rfcs: list[int] | None = None):
-    for ax in metadata_dict["axes"]:
-        if ax["unit"] is None:
-            ax.pop("unit")
+def _remove_none_values(obj: Any) -> Any:
+    """Recursively strip ``None``-valued keys from nested dicts and lists.
 
-        # Handle RFC 4: Remove orientation if RFC 4 is not enabled
-        if not is_rfc4_enabled(enabled_rfcs) and "orientation" in ax:
-            ax.pop("orientation")
+    In the OME-Zarr spec the absence of a field carries no significance: only
+    fields that are present are meaningful, and an explicit ``null`` is treated
+    as equivalent to omitting the field. Culling every ``None`` therefore keeps
+    the written metadata clean without enumerating optional fields per spec
+    version. Only ``None`` is removed; falsy-but-valid values such as ``0``,
+    ``0.0``, ``""`` and empty collections are preserved.
+    """
+    if isinstance(obj, dict):
+        return {
+            key: _remove_none_values(value)
+            for key, value in obj.items()
+            if value is not None
+        }
+    if isinstance(obj, list):
+        return [_remove_none_values(item) for item in obj]
+    return obj
 
-    if metadata_dict["coordinateTransformations"] is None:
-        metadata_dict.pop("coordinateTransformations")
 
-    if metadata_dict["omero"] is None:
-        metadata_dict.pop("omero")
+def _pop_metadata_optionals(
+    metadata_dict: dict, enabled_rfcs: list[int] | None = None
+) -> dict:
+    """Strip optional metadata fields that must not be serialized.
 
-    return metadata_dict
+    Removes the draft RFC 4 ``orientation`` from every axis unless RFC 4 is
+    enabled, then recursively culls all remaining ``None``-valued keys.
+    """
+    # RFC 4 anatomical orientation is a draft feature gated behind an explicit
+    # opt-in. Drop it from every axis (even when populated) unless RFC 4 is
+    # enabled; populated orientations on enabled writes survive the None
+    # stripping below.
+    if not is_rfc4_enabled(enabled_rfcs):
+        for ax in metadata_dict.get("axes", []):
+            ax.pop("orientation", None)
+
+    return _remove_none_values(metadata_dict)
 
 
 def _prep_for_to_zarr(store: StoreLike, arr: dask.array.Array) -> dask.array.Array:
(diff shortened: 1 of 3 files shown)


thewtex:
@jo-mueller please take a look :pray:

thewtex:
Addressed the review feedback in two focused commits:

**CodeRabbit nitpick — type hints on the helper** (9170f5e29)
Annotated `_remove_none_values(obj: Any) -> Any` and, for consistency, the
adjacent `_pop_metadata_optionals(metadata_dict: dict, ...) -> dict` in
`py/ngff_zarr/to_ngff_zarr.py`. No behavior change.

**Pre-merge check — docstring coverage** (cf162c423)
Added a docstring to `_pop_metadata_optionals` and to the test
functions/fixture in `py/test/test_metadata_none_removal.py`, matching the
existing one-line docstring convention already used in
`test_rfc4_integration.py`.

Both commits were reviewed with `coderabbit review --agent` (0 findings each),
are `ruff format`/`ruff check` clean, and the unit tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jo-mueller (approved):

jo-mueller:
LGTM! I'm not so familiar with the terminology RFC4 will introduce, would the presence of the `orientation` field (even when filled with `None`) carry meaning in that context?

thewtex:
@jo-mueller thanks for the review :pray: 

> the presence of the orientation field (even when filled with None) carry meaning in that context?

It should not -- I'll check that we are explicit about this in the specification during updates!

PR #561, opened 2026-06-23
feat: allow reading dev4 versions

jo-mueller:
@thewtex minor tweak in the wake of #98: Most datasets available on the web for testing write the version `0.6.dev4`, which is currently not supported to read, because the reader just checks the version against `version==0.6`. We can remove this later, for the time being this makes it easier to test things with other development versions.

Vice versa, I added a conditional check we'll also have to remove later, that will replace the written version `"0.6"` with `"0.6.dev4"`.

## Summary by CodeRabbit

* **New Features**
  * Added support for NGFF version **0.6.dev4** and made it the latest available version.
* **Bug Fixes**
  * Improved NGFF/OME-Zarr version detection so metadata parsing correctly handles **0.6** version variants.
  * When exporting NGFF **0.6**, the written root metadata is now normalized to **0.6.dev4** for consistency.

### DIFF
--- a/py/ngff_zarr/_supported_versions.py
+++ b/py/ngff_zarr/_supported_versions.py
@@ -29,7 +29,8 @@ class NgffVersion(StrEnum):
     V04 = "0.4"
     V05 = "0.5"
     V06 = "0.6"
-    LATEST = "0.6"
+    V06dev4 = "0.6.dev4"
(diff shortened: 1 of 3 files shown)


thewtex (approved):
@jo-mueller thanks!

PR #591, opened 2026-07-09
feat: non-destructive OME-Zarr spec-version upgrade (Python + TS + CLI)

### LINKED ISSUE #219: Prevent overwriting existing files
I tried to use ngff-zarr to convert a 0.4 image to 0.5. Accidentally, I used the same path for input and output. ngff-zarr did not fail and overwrote the original image. I am not sure if this is intended. I would have expected to get an error when trying overwrite existing files (without an `--overwrite` flag or similar).

thewtex:
## Summary

Adds a first-class **OME-Zarr specification-version upgrade** capability to both the Python (`ngff-zarr`) and TypeScript (`@fideus-labs/ngff-zarr`) packages, plus a `ngff-zarr upgrade` CLI subcommand and documentation. Upgrading a store to a newer OME-Zarr version (0.4 → 0.5 → 0.6) no longer requires an erase-then-reload round trip that risked data loss.

Fixes #219.

## Why

Previously the only way to move an existing OME-Zarr store to a newer spec version was to read it and write it back out, which erased the original before the rewrite completed — a data-loss hazard reported in #219. There was also no public API or CLI verb for a version upgrade at all. This PR provides a safe, explicit upgrade path with a **non-destructive default** and an in-place mode that leaves array chunk binaries byte-for-byte untouched.

### Python — `upgrade_ome_zarr()` / `upgrade_ngff_zarr()`
New module `py/ngff_zarr/upgrade_ome_zarr.py`, exported from `ngff_zarr`:

```python
upgrade_ome_zarr(input, output=None, *, version="0.6",
                 validate=False, storage_options=None, overwrite=True) -> None
```

Two modes:
- **In-place, metadata-only** (omit `output`): rewrites only root/array/group metadata; every array chunk is left byte-for-byte unchanged. For same-Zarr-format transitions (e.g. 0.5 ↔ 0.6) this is a pure metadata edit. Across the Zarr v2 ↔ v3 boundary (0.4 → 0.5/0.6) it writes Zarr v3 metadata with a `v2` `chunk_key_encoding` matching the source separator, so existing chunk binaries still resolve.
- **Write-to-new-store** (pass `output`): reuses the tested read/write pipeline for any supported transition; the source store is never mutated.

Input handling is hardened: source-version detection from root attributes, target-version validation, a same-version no-op short-circuit (bit-for-bit unchanged), and rejection of HCS-plate / bioformats2raw containers with per-image guidance. The in-place v3 → v2 downgrade is rejected with an actionable `ValueError` pointing to the write-to-new-store mode. `_write_root_ome_attrs()` was extracted from `_create_zarr_root()` in `to_ngff_zarr.py` so the writer and the upgrade path emit byte-identical root metadata.

### Python CLI — `ngff-zarr upgrade`
`cli.py`'s `main()` is refactored into a thin argv dispatcher: a leading `upgrade` token routes to a new `_upgrade_main()`; every other invocation runs the existing conversion parser (renamed `_convert_main`) unchanged, so the `ngff-zarr` entry point and all current flags behave identically. `upgrade` supports both modes (omit/pass `--output`), prints the detected source version / mode / target version via Rich, accepts `--to 0.4|0.5|0.6` (default 0.6), passes through `--validate` and `--overwrite/--no-overwrite`, and renders the documented errors as concise messages with a non-zero exit code (no traceback).

### TypeScript — `upgradeOmeZarr()`
Mirrors the Python behavior for parity. A shared, env-agnostic core (`upgrade_ome_zarr_common.ts`) is wrapped by thin Node (`upgrade_ome_zarr.ts`) and browser (`upgrade_ome_zarr-browser.ts`) entry points, exported from `mod.ts` and `browser-mod.ts`. In-place same-format upgrades rewrite only the root group's `zarr.json`; write-to-new-store reuses `fromOmeZarr` → `toOmeZarr` without mutating the source; cross-format in-place throws with guidance to pass `output`. A shared `buildRootAttributes()` helper was extracted so the Node/browser writers and the in-place rewrite share one source of truth for root metadata (accounting for the reductions in `to_ngff_zarr.ts` / `to_ngff_zarr-browser.ts`). In-place upgrade of a local *path* store is Node/Deno-only; `MemoryStore`/`FetchStore` inputs work everywhere.

### Docs & examples
- `docs/python.md`, `docs/cli.md`, `docs/typescript.md`: sections covering both modes, the array-chunks-untouched guarantee, and the cross-format constraints.
- `py/examples/upgrade_ome_zarr_example.py`: runnable, network-free example upgrading 0.4 → 0.6 (write-to-new-store) and 0.5 → 0.6 (in-place), asserting the in-place chunks stay byte-for-byte identical and the data reads back.

## Testing
- `py/test/test_upgrade_ome_zarr.py`: parametrized suite over the full (source, target) × (in_place, new_store) matrix, chunk byte-identity, OMERO/name/type/orientation round-trips, and edge cases (unsupported version, aliasing, HCS).
- `py/test/test_cli_upgrade.py`: network-free CLI tests for both modes, the rejected downgrade, invalid `--to`, `--help`, and subcommand dispatch.
- `ts/test/upgrade_ome_zarr_test.ts`: 8 Deno tests mirroring the Python matrix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added `ngff-zarr upgrade` CLI subcommand plus new upgrade APIs in Python (`upgrade_ome_zarr`) and TypeScript (`upgradeOmeZarr`), including browser support.
  * Supports safe in-place root-metadata upgrades and write-to-new-store upgrades for cross-format OME-Zarr version transitions.
* **Bug Fixes**
  * Prevents unsupported in-place cross-format upgrades and improves guidance for invalid targets/unsuitable inputs.
  * Preserves array/chunk payloads byte-for-byte during in-place metadata-only upgrades.
* **Tests**
  * Added end-to-end CLI and API test coverage for version transitions, no-op behavior, and on-disk immutability.
* **Documentation**
  * Expanded CLI/Python/TypeScript docs and added a runnable Python upgrade example.

### DIFF
--- a/docs/python.md
+++ b/docs/python.md
@@ -477,6 +477,76 @@ multiscales = from_ngff_zarr('cthead1.ome.zarr')
 to_ngff_zarr('cthead1_zarr2.ome.zarr', multiscales, version='0.4')
 ```
 
+## Upgrade OME-Zarr versions
+
+The conversion above re-reads and re-writes the entire pyramid. When you only
+need to change the *recorded specification version* of an existing store,
+[`upgrade_ome_zarr`] does it directly in one of two modes.
+
+**In-place, metadata-only.** When no `output` is given (or `output` resolves to
+the same store as `input`) and the source and target share the same underlying
+Zarr format -- for example 0.5 to 0.6, both Zarr v3 -- only the root group's
+`zarr.json` metadata is rewritten. **Every array chunk on disk is left
+byte-for-byte untouched**, avoiding the data loss of a naive "read, erase,
+re-write" upgrade. In-place upgrades that cross the Zarr v2/v3 boundary in the
+*upgrade* direction (0.4 to 0.5 or 0.4 to 0.6) are also metadata-only: each
+array's Zarr v3 `zarr.json` is given a `v2` chunk-key encoding so the existing
+chunk binaries resolve unchanged. The obsolete Zarr v2 sidecars (`.zarray`,
+`.zgroup`, `.zattrs`) are then removed, so the upgraded store is a valid Zarr v3
+/ OME-Zarr 0.5 (or 0.6) store *only* -- it is not simultaneously a valid Zarr v2
+/ OME-Zarr 0.4 store; only the chunk data binaries are reused, resolved through
+the `v2` chunk-key encoding. The reverse -- an in-place *downgrade* across that
+boundary (0.5/0.6 to 0.4) -- cannot preserve chunk keys and raises a
+`ValueError`; pass an `output` store instead.
+
+**Write-to-new-store.** When an `output` store distinct from `input` is given,
+the source is read lazily and re-written to `output` at the requested version
+through the standard write pipeline. Every supported transition (0.4, 0.5, 0.6,
+in either direction) works in this mode, and the source store is never erased.
+
+Upgrade a 0.5 store to 0.6 in place, keeping every array chunk:
+
+```python
+import numpy as np
+import ngff_zarr as nz
+
+# Synthesize a tiny store written at OME-Zarr 0.5 (Zarr v3).
+image = nz.to_ngff_image(np.zeros((4, 32, 32), dtype=np.uint8),
+                         dims=['z', 'y', 'x'])
+multiscales = nz.to_multiscales(image, scale_factors=[2])
+nz.to_ome_zarr('image.ome.zarr', multiscales, version='0.5')
+
+# Rewrite only the root metadata to 0.6; array chunks are left untouched.
+nz.upgrade_ome_zarr('image.ome.zarr', version='0.6')
+```
+
+Write an upgraded 0.6 copy from a 0.4 source, leaving the source intact:
+
+```python
+import numpy as np
+import ngff_zarr as nz
+
+# Synthesize a tiny store written at OME-Zarr 0.4 (Zarr v2).
+image = nz.to_ngff_image(np.zeros((4, 32, 32), dtype=np.uint8),
+                         dims=['z', 'y', 'x'])
+multiscales = nz.to_multiscales(image, scale_factors=[2])
+nz.to_ome_zarr('image_v04.ome.zarr', multiscales, version='0.4')
+
+# Write a new 0.6 store; 'image_v04.ome.zarr' is read lazily and never erased.
+nz.upgrade_ome_zarr('image_v04.ome.zarr', 'image_v06.ome.zarr', version='0.6')
+```
+
+Pass `validate=True` to validate the source metadata against the NGFF schema
+while reading, and `overwrite=False` (write-to-new-store only) to refuse to
+overwrite pre-existing data at `output`. A runnable version of both examples,
+which also proves the in-place upgrade leaves the array chunks byte-for-byte
+identical, is in
+[`py/examples/upgrade_ome_zarr_example.py`](https://github.com/fideus-labs/ngff-zarr/blob/main/py/examples/upgrade_ome_zarr_example.py).
+
+Note that `upgrade_ome_zarr` upgrades a single image. To upgrade an HCS plate or
+a bioformats2raw container, point at each contained image by its own path (for
+example `plate.ome.zarr/A/1/0`).
+
 [dataclass]: https://docs.python.org/3/library/dataclasses.html
 [dataclasses]: https://docs.python.org/3/library/dataclasses.html
 [Dask arrays]: https://docs.dask.org/en/stable/array.html
@@ -488,6 +558,7 @@ to_ngff_zarr('cthead1_zarr2.ome.zarr', multiscales, version='0.4')
 [`to_ngff_image`]: ./apidocs/ngff_zarr/ngff_zarr.to_ngff_image.md
 [`to_multiscales`]: ./apidocs/ngff_zarr/ngff_zarr.to_multiscales.md
 [`from_ngff_zarr`]: ./apidocs/ngff_zarr/ngff_zarr.from_ngff_zarr.md
+[`upgrade_ome_zarr`]: ./apidocs/ngff_zarr/ngff_zarr.upgrade_ome_zarr.md
 [`from_hcs_zarr`]: ./apidocs/ngff_zarr/ngff_zarr.hcs.md
 [Sharded Zarr]: https://zarr.dev/zeps/accepted/ZEP0002.html
 [tensorstore]: https://google.github.io/tensorstore/
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -752,6 +752,88 @@ const multiscalesV5 = await fromNgffZarr("data_v05.ome.zarr");
 await toNgffZarr("data_v04.ome.zarr", multiscalesV5, { version: "0.4" });
 ```
 
+### Upgrading OME-Zarr Store Versions
+
+The conversion above re-reads and re-writes the entire pyramid. To change only
+the *recorded specification version* of an existing store, `upgradeOmeZarr()`
+mirrors the Python `upgrade_ome_zarr` and offers the same two modes:
+
+```typescript
+function upgradeOmeZarr(
+  input: string | MemoryStore | FetchStore | Readable,
+  options?: {
+    output?: string | MemoryStore; // FetchStore is read-only, not a destination
+    version?: "0.4" | "0.5" | "0.6"; // default "0.6"
+    validate?: boolean;
+    overwrite?: boolean; // write-to-new-store only; default true
+  },
+): Promise<void>;
+```
+
+**In-place, metadata-only.** When `output` is omitted (or resolves to the same
+store as `input`) and the source and target share the same underlying Zarr
+format -- 0.5 to 0.6, both Zarr v3 -- only the root group's `zarr.json` is
+rewritten and every array chunk in the store is left byte-for-byte untouched,
+avoiding the data loss of a naive "read, erase, re-write" upgrade. An in-place
+upgrade that would cross the Zarr v2/v3 boundary
+(OME-Zarr 0.4 to 0.5/0.6) throws with guidance to pass `output` instead.
+
+```typescript
+import {
+  createAxis,
+  createDataset,
+  createMetadata,
+  createMultiscales,
+  createNgffImage,
+  toOmeZarr,
+  upgradeOmeZarr,
+  type MemoryStore,
+} from "@fideus-labs/ngff-zarr";
+
+// Build a small image and write it to a MemoryStore at version 0.5 (Zarr v3).
+const data = new Uint8Array(4 * 32 * 32);
+const image = await createNgffImage(
+  data.buffer,
+  [4, 32, 32],
+  "uint8",
+  ["z", "y", "x"],
+  { z: 1.0, y: 1.0, x: 1.0 },
+  { z: 0.0, y: 0.0, x: 0.0 },
+);
+const axes = [
+  createAxis("z", "space", "micrometer"),
+  createAxis("y", "space", "micrometer"),
+  createAxis("x", "space", "micrometer"),
+];
+const datasets = [createDataset("0", [1.0, 1.0, 1.0], [0.0, 0.0, 0.0])];
+const multiscales = createMultiscales([image], createMetadata(axes, datasets));
+
+const store: MemoryStore = new Map();
+await toOmeZarr(store, multiscales, { version: "0.5" });
+
+// Upgrade 0.5 -> 0.6 in place: only the root group's metadata is rewritten,
+// leaving every array chunk in the store untouched.
+await upgradeOmeZarr(store, { version: "0.6" });
+```
+
+**Write-to-new-store.** When `output` is a store distinct from `input`, the
+source is read lazily and re-written to `output` at the requested version
+through the standard write pipeline. Every supported transition (0.4, 0.5, 0.6,
+in either direction) works in this mode, and the source store is never mutated.
+
+```typescript
+import { upgradeOmeZarr, type MemoryStore } from "@fideus-labs/ngff-zarr";
+
+// `source` is an existing 0.4 store (e.g. a MemoryStore or FetchStore).
+const target: MemoryStore = new Map();
+await upgradeOmeZarr(source, { output: target, version: "0.6" });
+```
+
+In-place upgrade of a local **path** store is Node/Deno-only -- the browser build
+ships no filesystem store -- while `MemoryStore` inputs work in every
+environment, consistent with the other I/O functions. In the browser, use a
+`MemoryStore` for an in-place upgrade, or pass `output`.
+
 ### High Content Screening (HCS)
 
 Work with plate and well data:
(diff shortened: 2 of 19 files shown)


docs/typescript.md
```diff
@@ -752,6 +752,90 @@ const multiscalesV5 = await fromNgffZarr("data_v05.ome.zarr");
...
+const axes = [
+  createAxis("z", "space", "micrometer"),
+  createAxis("y", "space", "micrometer"),
+  createAxis("x", "space", "micrometer"),
+];
+const datasets = [createDataset("0", [1.0, 1.0, 1.0], [0.0, 0.0, 0.0])];
+const multiscales = createMultiscales([image], createMetadata(axes, datasets));
+
+const store: MemoryStore = new Map();
+await toNgffZarr(store, multiscales, { version: "0.5" });
```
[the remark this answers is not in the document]
thewtex: Fixed in 08bc01f01. `createNgffImage` is indeed `async` (it awaits `createTestZarrArray` and returns `Promise<NgffImage>`), so the example now does `const image = await createNgffImage(...)`. The surrounding snippet already runs under top-level await (`await toNgffZarr`, `await upgradeOmeZarr`), so this stays consistent.

Note: the same missing-`await` pattern exists in several **pre-existing** examples in this file (and the `#### createNgffImage()` reference block still shows it as sync-returning). Those predate this PR, so I scoped this fix to the example this PR introduced; flagging the rest as a possible follow-up docs pass.

py/ngff_zarr/upgrade_ome_zarr.py
```diff
@@ -0,0 +1,556 @@
...
+        try:
+            return zarr.open_group(open_store, mode="r", **kwargs).attrs.asdict()
+        except Exception:
+            return {}
+
+    root_attrs = _attrs()
+    if not root_attrs and zarr_version_major >= 3:
+        root_attrs = _attrs(zarr_format=2)
+
+    return root_attrs
```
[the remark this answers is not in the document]
thewtex: Fixed in 08bc01f01. The blanket `except Exception` in `_read_root_attrs` is now `except zarr.errors.GroupNotFoundError` — the same not-found signal already handled in `from_ngff_zarr.py`. That is precisely the exception raised when a Zarr v2 store is opened as v3, so the v0.4 → `zarr_format=2` fallback still works, while genuine failures propagate.

Verified empirically: a missing store now raises `FileNotFoundError` and corrupt root metadata raises `JSONDecodeError` (both were previously swallowed into a generic empty-attrs version-detection error), and v0.4/v0.5 detection is unchanged. Added `test_missing_store_surfaces_io_error` as a regression guard.

ts/src/io/upgrade_ome_zarr_common.ts
```diff
@@ -0,0 +1,165 @@
...
+    return true;
+  }
+  if (typeof input === "string" && typeof output === "string") {
+    // Compare local paths / URLs up to a trailing slash so `foo` and `foo/`
+    // match. Distinct string forms of the same directory are treated as
+    // different (and routed to the never-mutating write-to-new-store path).
+    return input.replace(/\/+$/, "") === output.replace(/\/+$/, "");
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
thewtex: Fixed in 08bc01f01. `isSameStore` now compares local path strings through a lexical `normalizeLocalPath` (collapses `//`, drops `.`, resolves `..`, strips a trailing slash), so `./data.ome.zarr` vs `data.ome.zarr` and `a/../data` are detected as the same store and routed to the safe in-place path — mirroring the Python side's `Path(...).resolve()` in `_stores_are_same`.

URLs keep the trailing-slash-only comparison (a leading `.`/`..` is meaningful in a URL path). The normalization is purely lexical, so it still can't see through symlinks or resolve a relative path against the cwd; those residual cases keep falling back to the never-mutating write-to-new-store path, so they remain safe by construction rather than risking an overwrite.

ts/src/io/upgrade_ome_zarr_common.ts
```diff
@@ -0,0 +1,165 @@
...
+  | MemoryStore
+  | zarr.FetchStore
+  | zarr.Readable;
+
+export interface UpgradeOmeZarrOptions {
+  /**
+   * Destination store or path. Omit (or pass the same store as `input`) for an
+   * in-place upgrade; pass a distinct store to write an upgraded copy there.
+   */
+  output?: string | MemoryStore | zarr.FetchStore;
```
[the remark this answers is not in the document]
thewtex: Fixed in 08bc01f01. Removed the read-only `zarr.FetchStore` from `UpgradeOmeZarrOptions.output` (now `string | MemoryStore`), since `toOmeZarr` rejects a `FetchStore` write destination at runtime. `FetchStore` remains valid as an `input` (reading), and the doc comment now says so explicitly.

docs/typescript.md
```diff
@@ -752,6 +752,90 @@ const multiscalesV5 = await fromNgffZarr("data_v05.ome.zarr");
...
+```typescript
+function upgradeOmeZarr(
+  input: string | MemoryStore | FetchStore | Readable,
+  options?: {
+    output?: string | MemoryStore | FetchStore;
+    version?: "0.4" | "0.5" | "0.6"; // default "0.6"
+    validate?: boolean;
+    overwrite?: boolean; // write-to-new-store only; default true
+  },
+): Promise<void>;
```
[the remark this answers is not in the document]
thewtex: Fixed in 08bc01f01. The docs `upgradeOmeZarr` signature now shows `output?: string | MemoryStore` with an inline note that `FetchStore` is read-only and not a destination. `input` still lists `FetchStore` because reading from it is valid. This matches the corresponding type change in `upgrade_ome_zarr_common.ts`.

py/ngff_zarr/upgrade_ome_zarr.py
```diff
@@ -0,0 +1,556 @@
...
+
+**In-place metadata-only rewrite.** When no `output` is given (or `output`
+resolves to the same store as `input`) and the source and target share the same
+underlying Zarr format -- for example 0.5 -> 0.6, both Zarr v3 -- only the root
+group's metadata (`zarr.json`) is rewritten. Every array chunk on disk is left
+byte-for-byte untouched. This is the fix for issue #219, where the previous
+"read, erase, re-write" approach destroyed the array data it was meant to
+preserve. In-place upgrades that would cross the Zarr v2<->v3 boundary
+(OME-Zarr 0.4<->0.5) are rejected with `NotImplementedError`; pass an `output`
+store instead.
```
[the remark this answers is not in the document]
thewtex: Fixed in 08bc01f01. Removed the stale sentence claiming in-place 0.4↔0.5 is rejected with `NotImplementedError`. As the following paragraph and the test suite already state, in-place **0.4 → 0.5/0.6 is supported** (metadata-only, via `_rewrite_v2_group_to_v3`, chunks left byte-for-byte untouched); only the cross-format **downgrade** (0.5/0.6 → 0.4) is rejected, with a `ValueError`. The docstring is now consistent with the implementation and the tests.

py/ngff_zarr/upgrade_ome_zarr.py
```diff
@@ -0,0 +1,556 @@
...
+
+    import asyncio
+
+    async def _prune() -> None:
+        keys = [key async for key in store.list()]
+        for key in keys:
+            if key.rsplit("/", 1)[-1] in sidecar_names:
+                await store.delete(key)
+
+    asyncio.run(_prune())
```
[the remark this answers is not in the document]
thewtex: Addressed in 08bc01f01. The `asyncio.run(_prune())` call — reached only for a non-filesystem store (e.g. a `MemoryStore`) during an in-place v2→v3 rewrite — is now wrapped: if it fails with a running-event-loop `RuntimeError` (as in a Jupyter notebook), it re-raises a clear, actionable message directing the user to pass an `output` store (the write-to-new-store path). Any other `RuntimeError` propagates unchanged.

This follows your suggested approach. A fully loop-aware async runner would be a broader change to make consistent with the codebase's shared `rfc9_zip` `asyncio.run` pattern, so I kept this to the clear-error improvement for now.

docs/python.md
```diff
@@ -516,6 +477,74 @@ multiscales = from_ngff_zarr('cthead1.ome.zarr')
...
+The conversion above re-reads and re-writes the entire pyramid. When you only
+need to change the *recorded specification version* of an existing store,
+[`upgrade_ome_zarr`] does it directly in one of two modes.
+
+**In-place, metadata-only.** When no `output` is given (or `output` resolves to
+the same store as `input`) and the source and target share the same underlying
+Zarr format -- for example 0.5 to 0.6, both Zarr v3 -- only the root group's
+`zarr.json` metadata is rewritten. **Every array chunk on disk is left
+byte-for-byte untouched.** This is the fix for
+[issue #219](https://github.com/fideus-labs/ngff-zarr/issues/219), where an
```
thewtex: rm github issues reference

docs/python.md
```diff
@@ -516,6 +477,74 @@ multiscales = from_ngff_zarr('cthead1.ome.zarr')
...
+earlier "read, erase, re-write" upgrade destroyed the array data it was meant to
+preserve. In-place upgrades that cross the Zarr v2/v3 boundary in the *upgrade*
+direction (0.4 to 0.5 or 0.4 to 0.6) are also metadata-only: each array's Zarr v3
+`zarr.json` is given a `v2` chunk-key encoding so the existing chunk binaries
+resolve unchanged. The reverse -- an in-place *downgrade* across that boundary
```
thewtex: clarify whether the result is a both a valid zarrv2 and zarrv3 , ome-zarr 0.4 and ome-zarr 0.5

docs/python.md
```diff
@@ -516,6 +477,74 @@ multiscales = from_ngff_zarr('cthead1.ome.zarr')
...
+
+```python
+import numpy as np
+import ngff_zarr as nz
+
+# Synthesize a tiny store written at OME-Zarr 0.5 (Zarr v3).
+image = nz.to_ngff_image(np.zeros((4, 32, 32), dtype=np.uint8),
+                         dims=['z', 'y', 'x'])
+multiscales = nz.to_multiscales(image, scale_factors=[2])
+nz.to_ngff_zarr('image.ome.zarr', multiscales, version='0.5')
```
thewtex: to_ngff_zarr -> to_ome_zarr

docs/python.md
```diff
@@ -516,6 +477,74 @@ multiscales = from_ngff_zarr('cthead1.ome.zarr')
...
+
+```python
+import numpy as np
+import ngff_zarr as nz
+
+# Synthesize a tiny store written at OME-Zarr 0.4 (Zarr v2).
+image = nz.to_ngff_image(np.zeros((4, 32, 32), dtype=np.uint8),
+                         dims=['z', 'y', 'x'])
+multiscales = nz.to_multiscales(image, scale_factors=[2])
+nz.to_ngff_zarr('image_v04.ome.zarr', multiscales, version='0.4')
```
thewtex: to_ngff_zarr -> to_ome_zarr

docs/typescript.md
```diff
@@ -794,6 +752,90 @@ const multiscalesV5 = await fromNgffZarr("data_v05.ome.zarr");
...
+```
+
+**In-place, metadata-only.** When `output` is omitted (or resolves to the same
+store as `input`) and the source and target share the same underlying Zarr
+format -- 0.5 to 0.6, both Zarr v3 -- only the root group's `zarr.json` is
+rewritten and every array chunk in the store is left byte-for-byte untouched.
+This is the fix for
+[issue #219](https://github.com/fideus-labs/ngff-zarr/issues/219), where a naive
```
thewtex: rm issues reference

docs/typescript.md
```diff
@@ -794,6 +752,90 @@ const multiscalesV5 = await fromNgffZarr("data_v05.ome.zarr");
...
+(OME-Zarr 0.4 to 0.5/0.6) throws with guidance to pass `output` instead.
+
+```typescript
+import {
+  createAxis,
+  createDataset,
+  createMetadata,
+  createMultiscales,
+  createNgffImage,
+  toNgffZarr,
```
thewtex: toNgffZarr -> toOmeZarr

docs/typescript.md
```diff
@@ -794,6 +752,90 @@ const multiscalesV5 = await fromNgffZarr("data_v05.ome.zarr");
...
+const axes = [
+  createAxis("z", "space", "micrometer"),
+  createAxis("y", "space", "micrometer"),
+  createAxis("x", "space", "micrometer"),
+];
+const datasets = [createDataset("0", [1.0, 1.0, 1.0], [0.0, 0.0, 0.0])];
+const multiscales = createMultiscales([image], createMetadata(axes, datasets));
+
+const store: MemoryStore = new Map();
+await toNgffZarr(store, multiscales, { version: "0.5" });
```
thewtex: toNgffZarr -> toOmeZarr

docs/python.md
thewtex: Done in 8589035de. Removed the `[issue #219](...)` hyperlink from the Python
upgrade section, keeping the rationale inline ("avoiding the data loss of a
naive read/erase/re-write upgrade") without the tracker link.

docs/python.md
thewtex: Clarified in 8589035de. The cross-format in-place upgrade is **not** dual-valid: after rewriting each array's Zarr v3 `zarr.json` (with a `v2` chunk-key encoding so the existing chunk binaries resolve), it removes the obsolete Zarr v2 sidecars (`.zarray`/`.zgroup`/`.zattrs`/`.zmetadata`). So a 0.4 -> 0.5/0.6 in-place upgrade yields a valid Zarr v3 / OME-Zarr 0.5 (or 0.6) store **only** — it is not simultaneously a valid Zarr v2 / OME-Zarr 0.4 store; only the chunk *data* binaries are reused.

Verified empirically: after an in-place 0.4->0.5, no v2 sidecars remain, opening the store with `zarr_format=2` raises `GroupNotFoundError`, and it reads back as validated OME-Zarr 0.5. The doc paragraph now states this explicitly (and `test_output_equal_to_input_cross_format` already asserts `_leftover_v2_sidecars == []`).

docs/python.md
thewtex: Done in 8589035de. Changed the in-place example to `nz.to_ome_zarr(...)`
(`to_ngff_zarr` is the backwards-compatible alias; `to_ome_zarr` is the
canonical name and is what the runnable example script already uses).

docs/python.md
thewtex: Done in 8589035de. Changed the write-to-new-store example's source-write call to
`nz.to_ome_zarr(...)` as well, for the same canonical-name reason.

docs/typescript.md
thewtex: Done in 8589035de. Removed the `[issue #219](...)` hyperlink from the TypeScript
upgrade section, keeping the inline rationale.

docs/typescript.md
thewtex: Done in 8589035de. Changed the import in the in-place example to `toOmeZarr`
(`toNgffZarr` is the alias; `toOmeZarr` is the canonical export).

docs/typescript.md
thewtex: Done in 8589035de. Changed the call site to `await toOmeZarr(store, multiscales, { version: "0.5" })` to match the updated import.

PR #594, opened 2026-07-09
fix(py): step region writes by the band width so the whole array is written

### LINKED ISSUE #592: Silent data loss: large-array (memory_target) write leaves part of the image unwritten
## Summary

On the large-array write path (used when `memory_usage(image) > config.memory_target`, i.e. images larger than roughly half of available RAM by default), `_compute_plane_regions` can leave part of the array unwritten. The unwritten region reads back as the fill value (zeros). There is no error and no warning about the missing data; the write reports success.

This is silent data loss in the current release (0.37.1), on the default path for large 3D images.

## Reproducer (self-contained)

```python
import numpy as np, dask.array as da
import ngff_zarr as nz

# 3D image; random data in 1..255 so any zero read back is unambiguously lost
data = np.random.default_rng(0).integers(1, 256, size=(1, 1, 40, 128, 256), dtype=np.uint8)
image = nz.to_ngff_image(da.from_array(data, chunks=16), dims=("t", "c", "z", "y", "x"))
ms = nz.to_multiscales(image, scale_factors=[1], chunks=16)

# Force the large-array path (a big image does this under the default memory_target)
nz.config.memory_target = 800_000
nz.to_ngff_zarr("loss.ome.zarr", ms, version="0.5", chunks_per_shard=2)

back = np.asarray(nz.from_ngff_zarr("loss.ome.zarr").images[0].data)
print("zeros in read-back:", int((back == 0).sum()), "of", back.size)  # 983040 of 1310720
print("byte-identical to source:", np.array_equal(back, data))         # False
```

On 0.37.1 this prints `zeros in read-back: 983040 of 1310720` and `byte-identical to source: False`: about 75% of the image is silently dropped. Writing the same image on the direct path (a large `memory_target`) is correct, and reading the store with plain `zarr` shows the same zeros, so the data is genuinely not on disk.

## Root cause

In `_compute_plane_regions` (`py/ngff_zarr/to_ngff_zarr.py`), the number of planes and strips is computed from the memory-band widths:

```python
num_y_splits = int(np.ceil(shape[y_index] / plane_slices))
...
num_x_splits = int(np.ceil(shape[x_index] / strip_slices))
```

but each write region is advanced by the chunk width, not the band width:

```python
region[y_index] = slice(plane_index * y_chunks, min((plane_index + 1) * y_chunks, arr.shape[y_index]))
region[x_index] = slice(strip_index * x_chunks, min((strip_index + 1) * x_chunks, arr.shape[x_index]))
```

When `strip_slices` (or `plane_slices`) is a multiple of the chunk width greater than one chunk, the loop runs `ceil(axis / band)` iterations of width `chunk`, so it covers only `ceil(axis / band) * chunk` and stops short of the axis. The tail beyond that is never written.

Example from the reproducer: axis `x = 256`, chunk 32, band `strip_slices = 128`. `num_x_splits = ceil(256 / 128) = 2`, stepped by 32, so only `x in [0, 64)` is written and `x in [64, 256)` is left as fill.

## Scope

- Affects the large-array / `memory_target` write path only. Small images and a large `memory_target` (the direct path) are unaffected.
- Requires the plane/strip subdivision, i.e. images with a `z` (or higher) axis. Pure 2D writes use a single region and are safe.
- Confirmed for both the default writer and the `use_tensorstore=True` writer: on the unpatched code both drop the same data, and both are repaired by the fix (they iterate the same regions from `_compute_write_regions`).
- Whether a given `(image, memory_target)` hits the gap depends on the region byte-sizing (`memory_target`, geometry, and dtype itemsize all feed the band-width arithmetic), so it is not tied to a single dtype; the reproducer above is deterministic. Reproduced across 14 of 96 swept 3D uint8 configurations, with losses up to 75%.

## Proposed fix

Advance each region by the band width the count already assumes:

```python
region[y_index] = slice(plane_index * plane_slices, min((plane_index + 1) * plane_slices, arr.shape[y_index]))
region[x_index] = slice(strip_index * strip_slices, min((strip_index + 1) * strip_slices, arr.shape[x_index]))
```

This makes the regions tile the whole array, with band sizes governed by `memory_target` as intended, so nothing is dropped. I have this implemented and verified: it closes all 14 of the 96 swept gaps (to 0), produces byte-identical read-back, leaves output unchanged for every case that already worked (backward-compatible), and passes the full test suite. Happy to open a PR.

## Environment

- ngff-zarr 0.37.1 (current `main`, 74cc620)
- zarr 3.2.1, dask 2026.7.0, Python 3.12

CraigMyles:
Fixes #592.

## Problem

On the large-array write path (`memory_usage(image) > config.memory_target`), `_compute_plane_regions` computes the number of planes and strips from the memory-band widths (`plane_slices`, `strip_slices`) but advances each write region by the chunk width (`y_chunks`, `x_chunks`). When a band is more than one chunk wide, the loop runs `ceil(axis / band)` iterations of `chunk` pixels each and stops short of the axis, leaving the tail unwritten. It reads back as the fill value (zeros), silently, with no error.

Minimal case: `x = 256`, chunk 32, band 128 gives `num_x_splits = 2` stepped by 32, covering only `x in [0, 64)`; `x in [64, 256)` is lost. See #592 for a runnable reproducer (about 75% of a 3D image dropped).

## Fix

Advance each region by the band width the loop count already assumes (`plane_slices` / `strip_slices` instead of `y_chunks` / `x_chunks`), at the three region-slice sites. The bands are already sized to `config.memory_target`, so the regions now tile the whole array within the intended memory budget.

## Backward compatibility

Output is unchanged for every case that already produced correct data; only previously-broken cases change (wrong to correct). Verified directly (same input, unpatched vs patched, compared to the source):

| Case | unpatched == source | patched == source | patched == unpatched |
|---|---|---|---|
| 2D (single region) | yes | yes | yes (no change) |
| 3D, direct path | yes | yes | yes (no change) |
| 3D region, previously correct | yes | yes | yes (no change) |
| 3D region, previously broken | no (data loss) | yes | no (fixed) |

The change is in version-agnostic tiling logic: no metadata, on-disk format, or spec field is touched, so OME-Zarr 0.4 and 0.5 behave identically. The `use_tensorstore=True` writer iterates the same regions and is repaired by the same change (confirmed).

## Tests

Adds `test_region_write_coverage.py`, which writes real data through the region path with a configuration that reproduces the loss on `main` and asserts byte-identical read-back. It fails on `main` and passes with this change.

## Validation

Full test suite on this branch: 727 passed, 3 skipped, no regressions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved region tiling for large array writes so coverage is complete, including the last rows/regions.
  * Ensured y-dimension partitioning matches the intended plane-based granularity for consistent write/read alignment.
* **Tests**
  * Added a regression test that exercises large-array region write round-trips across multiple shapes and memory target settings, with a guard for older Zarr versions.

### DIFF
--- a/py/ngff_zarr/to_ngff_zarr.py
+++ b/py/ngff_zarr/to_ngff_zarr.py
@@ -1251,12 +1251,12 @@ def _compute_plane_regions(
                     min((slab_index + 1) * z_chunks, arr.shape[z_index]),
                 )
                 region[y_index] = slice(
-                    plane_index * y_chunks,
-                    min((plane_index + 1) * y_chunks, arr.shape[y_index]),
+                    plane_index * plane_slices,
+                    min((plane_index + 1) * plane_slices, arr.shape[y_index]),
                 )
                 region[x_index] = slice(
-                    strip_index * x_chunks,
-                    min((strip_index + 1) * x_chunks, arr.shape[x_index]),
+                    strip_index * strip_slices,
+                    min((strip_index + 1) * strip_slices, arr.shape[x_index]),
                 )
                 plane_regions.append(tuple(region))
     else:
@@ -1268,8 +1268,8 @@ def _compute_plane_regions(
                 min((slab_index + 1) * z_chunks, arr.shape[z_index]),
             )
             region[y_index] = slice(
-                plane_index * y_chunks,
-                min((plane_index + 1) * y_chunks, arr.shape[y_index]),
+                plane_index * plane_slices,
+                min((plane_index + 1) * plane_slices, arr.shape[y_index]),
             )
             plane_regions.append(tuple(region))
 
--- a/py/test/test_region_write_coverage.py
+++ b/py/test/test_region_write_coverage.py
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: Copyright (c) Fideus Labs LLC
+# SPDX-License-Identifier: MIT
+import tempfile
+
+import dask.array as da
+import ngff_zarr as nz
+import numpy as np
+import pytest
+
+
+@pytest.mark.parametrize(
+    "dims, shape, chunks, mt",
+    [
+        (("t", "c", "z", "y", "x"), (1, 1, 40, 128, 256), 16, 800_000),
+        (("t", "c", "z", "y", "x"), (1, 1, 40, 160, 320), 16, 2_000_000),
+    ],
+)
+def test_large_array_region_write_covers_whole_array(dims, shape, chunks, mt):
+    """The large-array (memory_target) write path must write the whole array.
+
+    Regression for silent data loss: _compute_plane_regions counted planes and
+    strips by the memory-band width but stepped each region by the chunk width,
+    leaving the array tail unwritten (it read back as the fill value). Data is
+    in 1..255 so any zero read back is unambiguously lost.
+    """
+    data = np.random.default_rng(0).integers(1, 256, size=shape, dtype=np.uint8)
+    image = nz.to_ngff_image(da.from_array(data, chunks=chunks), dims=dims)
+    ms = nz.to_multiscales(image, scale_factors=[1], chunks=chunks)
+
+    old = nz.config.memory_target
+    nz.config.memory_target = mt
+    try:
+        store = tempfile.mkdtemp(suffix=".ome.zarr")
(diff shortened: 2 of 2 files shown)


thewtex (approved):
@CraigMyles thank you for the contribution! 🥇

PR #595, opened 2026-07-09
fix(py): clamp chunk/shard sizes to the axis instead of snapping to a divisor

### LINKED ISSUE #588: Large-array/sharded write collapses chunks to a tiny axis divisor (e.g. 5 px) for prime-factored dimensions
## Summary

`to_ngff_zarr` routes writing down a "large-array" / cached path (`_handle_large_array_writing`) whenever an image's array is large enough to cross `config.memory_target` (default = 50% of available RAM). On that path, chunk and shard sizes are chosen by `_find_optimal_chunk_size`, which **snaps each size to an exact integer divisor of the axis length** nearest the requested chunk. For axes whose length has a large prime factor, the nearest divisor is tiny, so the chunk collapses to a handful of pixels.

Real-world failure (converting whole-slide images to OME-Zarr v0.5): a `(t,c,z,y,x) = (1, 3, 1, 207634, 320095)` image, requested with 2048-px chunks, crossed `memory_target` and was written with **inner chunk `[1,1,1,1,5]` and shard `[1,3,1,14831,5]`**, i.e. 1x5-pixel chunks: on the order of 10^10 tiny chunks, ~6x storage inflation, and a multi-day write instead of ~20 minutes.

Why the divisors are tiny:
- `320095 = 5 x 64019` (64019 prime) -> divisors `{1, 5, 64019, 320095}`; the only one near a ~4096/8192 target is **5**.
- `207634 = 2 x 7 x 14831` (14831 prime) -> the near-target divisor is **14831**, whose divisors are `{1, 14831}`, so the inner chunk on that axis collapses to **1**.

The `min_divisor=16` in the function does **not** guard against this: it is only an *equal-distance tie-breaker*, never a lower bound, so `320095`'s nearest divisor (5) still wins.

The small/direct write path (below `memory_target`) is fine: it passes chunk sizes through untouched. The collapse only happens on the large-array path.

## Minimal reproducer (no large data needed)

A small synthetic array with a big-prime-factor axis, plus a very low `memory_target` to force the large-array path:

```python
import json, tempfile, numpy as np, dask.array as da
import ngff_zarr as nz

# x = 185 = 5*37  -> nearest divisor to the chunk target is 5
# y = 29 (prime)  -> inner chunk collapses to 1
arr = da.zeros((1, 3, 29, 185), dtype=np.uint8, chunks=(1, 1, 29, 185))
image = nz.to_ngff_image(arr, dims=("t", "c", "y", "x"))
ms = nz.to_multiscales(image, scale_factors=[2], chunks=8)     # intended inner chunk 8

nz.config.memory_target = 1     # tiny -> forces the large-array / cached write path
store = tempfile.mkdtemp(suffix=".ome.zarr")
nz.to_ngff_zarr(store, ms, version="0.5", chunks_per_shard=2)   # intended shard 16

root = json.load(open(f"{store}/zarr.json"))
l0 = root["attributes"]["ome"]["multiscales"][0]["datasets"][0]["path"]
meta = json.load(open(f"{store}/{l0}/zarr.json"))
print("requested inner: 8, shard: 16")
print("got  inner:", meta["codecs"][0]["configuration"]["chunk_shape"],
      " shard:", meta["chunk_grid"]["configuration"]["chunk_shape"])
```

**Expected:** inner `[1, 3, 8, 8]`, shard `[1, 3, 16, 16]` (the requested sizes, clamped to the axis where smaller).

**Actual (`main`, v0.37.1):**
```
requested inner: 8, shard: 16
got  inner: [1, 3, 1, 5]  shard: [1, 3, 29, 5]
```
The x axis (185 = 5x37) collapses to a 5-px chunk; the prime y axis (29) collapses to a 1-px inner chunk. Raising `memory_target` above the array size (so the direct path runs instead) gives the correct `8`/`16`. Toggling only the memory branch flips buggy to correct.

## Root cause

`_find_optimal_chunk_size(first_chunk, dim_size, min_divisor=16)`, in `py/ngff_zarr/to_ngff_zarr.py` and duplicated in `py/ngff_zarr/to_multiscales.py`, only ever returns an exact divisor of `dim_size`. Zarr v3 permits a **partial final chunk**, so chunk/shard sizes need not divide the axis evenly (the only place divisibility is genuinely required is the sharding *inner-chunk-divides-shard* constraint). Targeting the requested size and clamping to the axis length, instead of snapping to a divisor, fixes it.

## Environment

- `main` @ `74cc620` (v0.37.1); zarr 3.2.1, dask 2026.x, Python 3.12. Reproduces on both aarch64 and x86_64.

## Existing issues

Searched open + closed issues/PRs. None address the degenerate-chunk / prime-dimension collapse.

## Scope

Specific to the Python implementation. The TypeScript implementation (`ts/`) passes chunk sizes through unchanged and doesn't support sharding (it defers sharding to Python), so it's unaffected.

---

*I have a tested fix ready (replaces the divisor-snapping with a clamp, plus a small helper for the one spot zarr genuinely needs divisibility, the sharding inner chunk, falling back to the whole shard for a prime shard rather than a 1-px chunk). It passes the existing test suite. Happy to open a PR if this approach looks right.*

CraigMyles:
Fixes #588.

> **Depends on #594** (the region-write data-loss fix) and is stacked on it. Please review and merge #594 first. This PR is a draft until then: its diff currently also includes #594's change, and the moment #594 merges I will rebase this branch onto `main` so it becomes clamp-only and mark it ready for review.

## Problem

The large-array / sharded write path chose sizes with `_find_optimal_chunk_size`, which snapped each chunk and shard size to an exact divisor of the axis length. For an axis with a large prime factor the nearest divisor is tiny, so the chunk collapsed to a few pixels. A real `(1, 3, 1, 207634, 320095)` slide was written with inner chunk `[1,1,1,1,5]` / shard `[1,3,1,14831,5]`: about 1e10 tiny chunks and roughly 6x storage inflation. `min_divisor=16` is only a tie-breaker, not a lower bound.

## Fix

Zarr v3 permits a partial final chunk and a partial final shard, so sizes need not divide the axis. Replace the divisor-snapping helper with a clamp (`max(1, min(target, axis))`) in both `to_ngff_zarr.py` and `to_multiscales.py`, and add `_inner_chunk_for_shard` for the one place divisibility is required (the sharding inner chunk must divide the shard): it returns the largest chunk `<=` the target that divides the shard, never a 1-px chunk.

## Result

On the #588 reproducer: before `inner [1,3,1,5] / shard [1,3,29,5]`; after `inner [1,3,8,8] / shard [1,3,16,16]`.

## Tests

Adds `test_degenerate_chunk_regression.py`: prime axes no longer collapse, a partial final shard reads back unchanged, a user-supplied Zstd level is preserved, and a realistic multi-scale write reads back byte-identical.

## Note on a dask PerformanceWarning

Allowing a partial final shard makes dask emit its region-alignment `PerformanceWarning` on this path (`main` does not, because divisor-snapping kept writes aligned at the cost of the collapse). With the region-write fix underneath, the write is correct (verified byte-identical); the warning is a known dask false positive on the region path (dask PR #12161).

## Validation

Full test suite: 731 passed, 3 skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NGFF OME-Zarr v0.5 large-array and multiscale writing to avoid chunk/shard sizing collapsing on prime-sized dimensions.
  * Updated chunk sizing to allow partial final chunks without enforcing even divisibility, improving correctness on spatial axes.
  * Preserved user-configured Zstd compression settings and ensured byte-identical multiscale round-trips.
* **Tests**
  * Added regression tests covering prime/2×prime axes, partial final shard reads, and codec preservation.
* **Chores**
  * Updated the supported Dask version to 2026.1.2 and refreshed the CI Dask pin.

### DIFF
--- a/.github/workflows/python_dask.yaml
+++ b/.github/workflows/python_dask.yaml
@@ -12,7 +12,7 @@ jobs:
         include:
           - python-version: "3.12"
             zarr-version: ">=3.0.0"
-            dask-version: "==2025.11.0"
+            dask-version: "==2026.1.2"
     steps:
       - uses: actions/checkout@v4
 
--- a/py/ngff_zarr/to_multiscales.py
+++ b/py/ngff_zarr/to_multiscales.py
@@ -171,11 +171,11 @@ def _large_image_serialization(
         path = f"{base_path}/slabs"
         slabs = data.rechunk(rechunks)
 
-        # For spatial dimensions, ensure Zarr chunk sizes are divisors of
-        # the dimension sizes so older dask versions (e.g. 2025.11.0) can
-        # write safely with regions.  For non-spatial dims (t, c), use the
-        # slab chunk directly to keep channels / timepoints together and
-        # avoid dask rechunking (gh-issue-487).
+        # For spatial dimensions, clamp the Zarr chunk size to the axis length.
+        # Zarr v3 allows a partial final chunk, so the region write stays safe
+        # without snapping to a divisor.  For non-spatial dims (t, c), use the
+        # slab chunk directly to keep channels / timepoints together and avoid
+        # dask rechunking (gh-issue-487).
         chunks = tuple(
             _find_optimal_chunk_size(c[0], data.shape[i])
             if dim in _spatial_dims
--- a/py/ngff_zarr/to_multiscales.py
+++ b/py/ngff_zarr/to_multiscales.py
@@ -91,45 +91,16 @@ def _ngff_image_scale_factors(ngff_image, min_length, out_chunks):
     return scale_factors
 
 
-def _find_optimal_chunk_size(first_chunk, dim_size, min_divisor=16):
-    """Find a chunk size that divides evenly into dim_size and is ideally divisible by min_divisor.
+def _find_optimal_chunk_size(first_chunk, dim_size):
+    """Target the requested chunk size, clamped to the axis length.
 
-    The returned chunk size will:
-    1. Divide evenly into dim_size (required for safe region writes)
-    2. Be as close as possible to first_chunk
-    3. Preferably be divisible by min_divisor for performance
+    Zarr v3 permits a partial final chunk, so the cache chunk size need not
+    divide the dimension evenly (region slabs are chunk-aligned and the final
+    slab is truncated to the axis, so a partial final chunk is safe). Do not
+    snap to an axis divisor: for dimensions with large prime factors the
+    nearest divisor collapses to a tiny chunk (e.g. 320095 = 5 * 64019 -> 5).
     """
-    # If dimension is very small, just use it directly
-    if dim_size <= min_divisor:
-        return dim_size
-
-    # Start with the target chunk size
-    target = first_chunk
-
-    # First try to find a divisor of dim_size that's divisible by min_divisor
-    # and close to our target
-    best_chunk = dim_size  # Fallback: use full dimension
-    best_distance = abs(dim_size - target)
-
-    # Check all divisors of dim_size
-    for i in range(1, int(np.sqrt(dim_size)) + 1):
-        if dim_size % i == 0:
-            # i and dim_size//i are both divisors
-            for candidate in [i, dim_size // i]:
-                distance = abs(candidate - target)
-                # Prefer divisors that are multiples of min_divisor
-                is_multiple = candidate % min_divisor == 0
-
-                # Update if closer to target, with preference for multiples of min_divisor
-                if distance < best_distance or (
-                    distance == best_distance
-                    and is_multiple
-                    and best_chunk % min_divisor != 0
-                ):
-                    best_chunk = candidate
-                    best_distance = distance
-
-    return best_chunk
+    return max(1, min(int(first_chunk), int(dim_size)))
 
 
 def _large_image_serialization(
--- a/py/ngff_zarr/to_ngff_zarr.py
+++ b/py/ngff_zarr/to_ngff_zarr.py
@@ -844,48 +844,34 @@ def _handle_large_array_writing(
         else:
             shrink_factors.append(1)
 
-    # Ensure chunks are compatible with Dask's to_zarr when writing with regions.
-    # The Zarr chunk size must divide evenly into the dimension size to avoid
-    # PerformanceWarning and potential data loss during region writes.
-    def _find_optimal_chunk_size(first_chunk, dim_size, min_divisor=16):
-        """Find a chunk size that divides evenly into dim_size and is ideally divisible by min_divisor.
+    # Region writes and shard/array chunk sizes do not need to divide the
+    # dimension evenly: Zarr v3 permits a partial final chunk. Region slabs are
+    # chunk-aligned and the final slab is truncated to the axis, so a partial
+    # final chunk is safe.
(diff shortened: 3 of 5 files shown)


CraigMyles:
@thewtex thanks for merging #594, and for the earlier thumbs-up on the clamp approach!

This is rebased to clamp-only on top of `main` now. CI is green except one job: `test (3.12, >=3.0.0, ==2025.11.0)`.

The clamp deliberately allows a partial final shard (spec-legal in Zarr v3 / OME-Zarr), so the large-array writes are no longer chunk-aligned and `dask.array.to_zarr` emits its `PerformanceWarning`. On current dask that is just a warning, everything passes, and the output is byte-identical, so I don't think the warning itself is worth engineering around. The only failure is on dask 2025.11.0, where that warning is *raised as an exception* (the behaviour dask fixed in #12161), which trips the two writes that go through `dask.array.to_zarr`: the region write in `to_ngff_zarr.py` and the slab-cache write in `to_multiscales.py`.

This looks like a dask bug on one pinned version rather than something the write code should bend around, so I'd handle it at the dependency level, bumping the dask floor past the buggy version or dropping/relaxing that pinned config, and leave the write path and the (legitimate) warning as they are.

If you'd rather keep that dask version green and silence it in code, the next best options are either:
- write each region synchronously (`zarr_array[region] = arr.compute()`) at both sites, which is byte-identical and still memory-bounded but drops the parallel write there, or
- a scoped suppression of just that `PerformanceWarning` around the two writes.

thewtex:
@CraigMyles yes, we could update dask dependency version. One constraint is our Python version 3.10 support -- if we can keep it, let's do it. We recently dropped Python 3.9 support, so that may help.

CraigMyles:
@thewtex thanks. Python 3.10 is safe to keep: every dask release from 2025.11.0 through the current 2026.7.0 declares `requires-python >=3.10`, so a floor bump doesn't touch 3.10.

I reproduced the failing job locally to find the right floor, and it isn't only 2025.11.0. Running the two regression test files on Python 3.12 / zarr 3.2.1:

- dask 2025.10.0: the partial-shard tests fail on a wrong read-back. The write proceeds and loses data, which is the case the warning warns about.
- dask 2025.11.0: the same writes raise the `PerformanceWarning` as an exception (the #12161 behaviour), 8 tests.
- dask 2026.1.2: all pass, warning is benign.
- dask 2026.7.0 (latest): pass.

Since 2025.12.0, 2026.1.0 and 2026.1.1 are already excluded here, 2026.1.2 is the lowest usable release. So I bumped the floor rather than adding another exclusion:

- `py/pyproject.toml`: `dask[array]>=2026.1.2` (this supersedes the existing `!=2025.12.0,!=2026.1.0,!=2026.1.1`, all below it).
- `python_dask.yaml`: the compat pin `2025.11.0` -> `2026.1.2`.

This keeps 3.10 and leaves the write path and the legitimate warning as they are. Pushed to the PR.

thewtex:
@CraigMyles thanks for testing. `2026.1.2` min sounds good. :+1:

thewtex:
@CraigMyles thanks for the update. Please see agent review comments.

thewtex (approved):
@CraigMyles thank you! 🥇

PR #604, opened 2026-07-21
fix: reject RFC-4 orientation on non-spatial axes and duplicate anatomical axes

vboussot:
## Problem

`validate_rfc4_orientation` accepts two malformed inputs that RFC-4 forbids:

1. **Orientation on a non-spatial axis** — RFC-4 orientation is only defined for `space` axes, yet an orientation declared on a `time`/`channel` axis was silently accepted.
2. **Duplicate anatomical axis** — RFC-4 directions are mutually exclusive (one per antonym pair), yet two spatial axes on the same axis (e.g. `left-to-right` and `right-to-left`) were silently accepted.

## Fix

Two checks added to `validate_rfc4_orientation`, surfaced through the existing `SpecRule` channel as `axis-orientation-on-non-space` and `axis-orientation-unique-axis`. Well-formed metadata is unaffected; only these two malformed shapes are now rejected.

## Tests

- unit tests for both new rules, at the function level and via the `validate_axis_orientation` → `SpecRule` mapping;
- the message-marker stability test updated with the two new markers.

All RFC-4 orientation tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened RFC 4 axis-orientation validation by rejecting any `orientation` on non-space axes (including `orientation: {}`).
  * Added enforcement to prevent multiple spatial axes from resolving to the same anatomical axis, avoiding conflicting/duplicate direction metadata.
  * Validation now reports dedicated rule identifiers for these cases (Python + TypeScript).

* **Tests**
  * Expanded unit and message-mapping tests to cover non-space rejection, empty-orientation rejection, and duplicate anatomical-axis (including same-name collision) failures.

### DIFF
--- a/py/ngff_zarr/rfc4_validation.py
+++ b/py/ngff_zarr/rfc4_validation.py
@@ -23,6 +23,37 @@ def load_rfc4_orientation_schema() -> dict:
     return json.loads(schema)
 
 
+# Each RFC 4 value maps to a stable key for the anatomical axis it lies on, so the
+# two antonyms of a pair (e.g. "left-to-right" / "right-to-left") collapse to one
+# key. Used to enforce "one direction per anatomical axis" (mutual exclusion).
+_ANATOMICAL_AXIS_OF: dict[str, str] = {
+    "left-to-right": "left-right",
+    "right-to-left": "left-right",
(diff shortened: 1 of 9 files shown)


thewtex (approved):
👍

PR #605, opened 2026-07-21
feat(py): add `ngff-zarr conformance` RFC-4 verdict subcommand

vboussot:
## What

Adds a `conformance` subcommand to the Python CLI:

```
ngff-zarr conformance <input>
```

It reads one OME-Zarr and prints a single canonical JSON verdict:

```json
{"input": "...", "format": "ome-zarr", "rfc4_valid": true,
 "axes": {"z": "inferior-to-superior", ...}, "violations": [], "warnings": []}
```

## Why

This is the CLI a tool-agnostic RFC-4 conformance driver calls per input and diffs
against its manifest, so ngff-zarr can be checked as a *verified* RFC-4 validator.

## Notes

- Reuses `validate_rfc4_orientation` (no rule reimplemented) and reads the raw axes
  so malformed metadata is still classifiable (the high-level reader would reject
  it before it could be reported).
- Python CLI only; there is no TypeScript CLI, so no parity counterpart is needed.
- Builds on #604 (the RFC-4 validation completeness fix); merge that first so the
  subcommand rejects the non-space / duplicate-axis cases and passes the suite.

## Summary by CodeRabbit

- **New Features**
  - Added an `ngff-zarr conformance` command that outputs RFC 4 OME-Zarr conformance as JSON, including derived axis-orientation mapping and structured violation codes.
  - Supports OME-Zarr directories, v2-style `.zattrs`, and `.zip` archives.
- **Bug Fixes**
  - RFC 4 validation now flags `orientation` on non-`space` axes and detects conflicts where multiple spatial axes map to the same anatomical axis.
  - Malformed or incomplete metadata yields consistent “invalid” conformance output.
- **Tests**
  - Added CLI integration and extensive RFC 4/structural validation tests covering valid cases and edge-case failures.

### DIFF
--- a/py/ngff_zarr/cli.py
+++ b/py/ngff_zarr/cli.py
@@ -469,6 +469,19 @@ def _ngff_image_to_multiscales(
 
 
 def main():
+    # RFC 4 conformance mode: `ngff-zarr conformance <input>` prints one canonical
+    # JSON verdict per input, for a tool-agnostic conformance driver to diff.
+    if len(sys.argv) >= 2 and sys.argv[1] == "conformance":
+        import json
+
+        from .rfc4_conformance import conformance_report
+
+        if len(sys.argv) != 3:
+            print("usage: ngff-zarr conformance <input>", file=sys.stderr)
+            raise SystemExit(2)
+        print(json.dumps(conformance_report(sys.argv[2]), indent=2))
+        return
+
     parser = argparse.ArgumentParser(
         description="Convert datasets to and from the OME-Zarr Next Generation File Format.",
         formatter_class=RichHelpFormatter,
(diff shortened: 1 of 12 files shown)


vboussot:
Closing in favour of #608, which now carries this commit directly instead of stacking on top of it.

While building the conformance driver for this subcommand, the suite surfaced three ways `validate_rfc4_orientation` diverges from the current RFC 4 text — an all-or-none completeness rule the specification does not state, `null` treated as invalid rather than as equivalent to absent, and an orientation `type` checked only for internal consistency rather than against the single defined value. Those fixes are inseparable from the verdict this subcommand emits, so keeping them in one PR is easier to review than two stacked ones.

Nothing is lost: #608 is rebased on `main` after #604 merged and contains this commit unchanged, plus the specification fixes. The OME-Zarr conformance cases go from 7/24 to 24/24.

Issue #87, opened 2024-08-22
CI doesn't run any tests


btbest:
Not sure this is intentional :)

![image](https://github.com/user-attachments/assets/a3115997-9e40-41a2-9c70-e663c272ec57)

thewtex:
:open_mouth: WIP in #90 #89 

Tests should pass locally.

thewtex:
Addressed in `main`

PR #611, opened 2026-07-23
fix(py,ts): support OME-Zarr RFC-3 (arbitrary axis dimensions, names, types, ordering)

### LINKED ISSUE #599: Add support for RFC-3
RFC-3 has been updated (ome/ngff#560) to add axis naming recommendations, be more specific about changes to the spec document, and provide sample data. It would be nice to be able to use ngff-zarr or a branch thereof to read and write RFC-3 compliant data.

vboussot:
Closes #599

Adds RFC-3 support (expanded axis count, names, types and order), enabled only
for OME-Zarr **0.9.dev1**. Versions 0.4, 0.5 and 0.6 keep their axis limits,
since their schemas cap axes at 5 and require 2 or 3 `space` axes.

`0.9.dev1` is the version string proposed in [ome/ngff-spec#190](https://github.com/ome/ngff-spec/pull/190),
an open WIP that describes it as 0.6rc1 plus RFC-3, RFC-4 and probably
[#87](https://github.com/ome/ngff-spec/issues/87); the draft currently carries
RFC-3 only. Glencoe Software writes the same string in
[bioformats2raw#330](https://github.com/glencoesoftware/bioformats2raw/pull/330).

RFC-4 needs no work here: `v09` re-exports the v0.6 model except for `Axis`,
`CoordinateSystem` and `Metadata`, which it widens, so anatomical orientation
already round-trips at that version.

## How it works

- **In memory**: `NgffImage` and `Metadata` carry no axis restriction.
- **Writing**: refused if the target version serializes an axis model it cannot
  express.
- **Reading**: unchanged, still permissive; validation applies the axis rules of
  the version the store declares.

```python
to_ome_zarr("out.zarr", six_axis_image, version="0.5")
# ValueError: Cannot write OME-Zarr version="0.5" ... Pass version="0.9.dev1".

to_ome_zarr("out.zarr", six_axis_image, version="0.9.dev1")   # ok
```

The gate sees exactly what the target version writes. A v0.4/v0.5 target
serializes one flat `axes` and drops the coordinate systems, so a system the
downgrade discards is not gated: Python reaches that set by gating after
`Metadata.to_version`, TypeScript by checking the target version in `axisViews`.
Both refusal messages are byte-identical and pinned as the same literal in both
suites.

`0.9.dev1` is opt-in. `LATEST` stays `0.6.dev4`, so nothing changes for existing
callers. OME publishes no 0.9.dev1 JSON Schema yet, so none is bundled and
`validate=True` says so rather than failing obscurely. `ngff-zarr upgrade --to
0.9.dev1` is accepted alongside the library API.

## One bug fixed

An axis with no `type` is rejected on read, although the 0.4 schema allows it.

## Four things to flag

**Writing now applies the axis rules.** The writer applied none before this PR,
so a 6-axis model serialized happily at 0.5. That is the point of the gate, but
it is a behaviour change for anyone deliberately writing such stores, and the
only way to keep writing them is `version="0.9.dev1"`.

**The space-axis floor is new below 0.9.dev1.** Only the ceiling of 3 `space`
axes was enforced; the 0.4 schema also states `minContains: 2`, and the upstream
conformance corpus files `one_space_axes` under `invalid/`. `validate_structural`
and the writer now reject a single `space` axis at 0.4, 0.5 and 0.6. The v0.6
`axes` schema is a `oneOf`, so an RFC-5 `array` coordinate system, which carries
no `space` axis, is exempt from that floor (the 3-axis cap and the `(z, y, x)`
suffix check still apply).

**A non-canonical axis class order is refused.** Axis order is a spec MUST and
`validate_structural` rejects it, so the writer rejects it too and the two
finally agree. #623 normalizes generated axes to `(t, c, z, y, x)`, so the
Python pipeline produces a canonical order and a model that reaches the gate out
of order comes from the caller.

That normalization is Python only: the TypeScript port has no equivalent, and no
lazy transpose to build one on, so a model Python normalizes is refused by the
TypeScript writer below 0.9.dev1. Worth its own issue.

**The ingestion path does not accept an RFC-3 axis model yet.** `to_ngff_image`
caps `dims` at the `{c, x, y, z, t}` vocabulary and 5 axes, and `to_multiscales`
raises on any other name, so an RFC-3 model enters the library through
`from_ome_zarr` or a hand-built `Metadata` rather than through the array
ingestion helpers. Also worth its own issue.

## On `axis-names-unique`

The new rule states RFC-3 rule 5, "axis names MUST NOT be repeated within a
dataset". No released schema carries it: 0.4 and 0.5 say nothing, and 0.6 has
only a non-normative `description`. Applying it below 0.9.dev1 is therefore a
strictness choice rather than a spec MUST of those versions, and the rule
reference says so. The case-insensitivity half of rule 5 is a SHOULD and is not
enforced.

The rule's position in the fail-fast cascade is locked in both ports: the shared
`EXPECTED_EVALUATION_ORDER` now runs eleven stages with `axis-names-unique` at
position 5, and moving the call in either direction fails the parity suite.

## Tests

Python 975 passed on 3.14 with zarr-python 3, 679 passed on 3.10 with
zarr-python 2, and 572 passed in TypeScript. Each commit is green on the axis,
parity, RFC-3 and CLI suites in both languages.

The 3.10 environment also reports 30 failures in
`test_itk_transform_resample_bounding_box.py`, all
`ImportError: cannot import name 'resample_bounding_box' from
'itkwasm_downsample'`. That file and the ITK code paths are untouched here.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in OME-Zarr 0.9.dev1 support for Python and TypeScript reading, writing, and upgrades.
  * Added flexible RFC-3 axes with arbitrary names, types, units, ordering, and dimensionality.
  * Added RFC-5 map-axis, by-dimension, and bijection transformations.
  * Added v0.6 array coordinate systems and version-aware validation.
  * Added duplicate axis-name detection and clearer validation details.
  * Expanded remote and S3 storage support, including anonymous access defaults.

* **Documentation**
  * Updated API, CLI, Python, and TypeScript guidance.

* **Tests**
  * Expanded coverage for validation, transformations, round trips, and upgrades.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ The main Python package provides:
 - Optional OME-Zarr data model validation during reading
 - Writes OME-Zarr v0.4 to v0.6
 - v0.6 adds RFC-5 coordinate systems and transformations
+- Reads and writes the opt-in development version `0.9.dev1`, which adopts RFC-3
+  (expanded axis count, names, types and order)
 - [Sharded Zarr] stores
 - Optional writing via zarr-python 2, zarr-python 3, [zarrista] or zarrita (TypeScript)
 - [Anatomical orientation metadata](./docs/rfc4.md) (RFC-4)
@@ -67,7 +69,7 @@ The TypeScript package provides universal OME-Zarr support for modern JavaScript
 - 🦕 **Deno-first** with first-class TypeScript support
 - 📦 **Universal compatibility** - Works in Deno, Node.js, and browsers
 - 🔍 **Type-safe** with Zod schema validation
-- 🗂️ **OME-Zarr v0.4, v0.5, and v0.6** support (v0.6 adds RFC-5 coordinate systems and transformations)
+- 🗂️ **OME-Zarr v0.4, v0.5, and v0.6** support (v0.6 adds RFC-5 coordinate systems and transformations), plus the opt-in development version `0.9.dev1` which adopts RFC-3 (expanded axis count, names, types and order)
 - 🌐 **Web ready** - No filesystem dependencies, works with remote stores
 - 🏗️ **Mirrors Python API** - Familiar interfaces for Python users
 - 📚 **Lazy loading** - Efficient handling of large datasets
--- a/docs/python.md
+++ b/docs/python.md
@@ -213,7 +213,7 @@ To read an OME-Zarr file, use [`from_ngff_zarr`], which returns the
 >>> multiscales = nz.from_ngff_zarr('cthead1.ome.zarr')
 ```
 
-OME-Zarr version 0.1 to 0.6 is supported. Version 0.6 adds RFC-5 coordinate systems and transformations.
+OME-Zarr version 0.1 to 0.6 is supported. Version 0.6 adds RFC-5 coordinate systems and transformations. The opt-in development version `0.9.dev1` additionally adopts RFC-3, that expands supported axis counts, names, types and order.
 
 The `store` argument accepts:
 
@@ -551,8 +551,10 @@ boundary (0.5/0.6 to 0.4) -- cannot preserve chunk keys and raises a
 
 **Write-to-new-store.** When an `output` store distinct from `input` is given,
 the source is read lazily and re-written to `output` at the requested version
-through the standard write pipeline. Every supported transition (0.4, 0.5, 0.6,
-in either direction) works in this mode, and the source store is never erased.
+through the standard write pipeline. Every transition among 0.4, 0.5, 0.6 and
+0.9.dev1 works in this mode whenever the target version can express the axis
+model, and the source store is never erased. An RFC-3 axis model is refused
+below 0.9.dev1, so a store using one only converts upward.
 
 Upgrade a 0.5 store to 0.6 in place, keeping every array chunk:
 
--- a/docs/spec_features.md
+++ b/docs/spec_features.md
@@ -38,6 +38,11 @@ supported by `ngff-zarr`.
 - **OME-Zarr v0.4 to v0.6**: Writes OME-Zarr versions 0.4 to 0.6, including
   RFC-4 anatomical orientation. v0.6 additionally adds RFC-5 coordinate systems
   and transformations.
+- **OME-Zarr 0.9.dev1**: Reads and writes the development version that adopts
+  RFC-3, which extends support for the number, names, types and order
+  of axes. It is opt-in: pass `version="0.9.dev1"` explicitly. The default
+  target is unchanged, and OME publishes no JSON Schema for it yet, so schema
+  validation is unavailable at that version.
 
 ## High Content Screening (HCS)
 
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -9,7 +9,7 @@ NGFF-Zarr provides a TypeScript implementation for working with OME-Zarr data st
 - 🦕 **Deno-first**: Built for Deno with first-class TypeScript support
 - 📦 **Universal compatibility**: Works in Deno, Node.js, and browsers
 - 🔍 **Type-safe**: Full TypeScript support with Zod schema validation
-- 🗂️ **OME-Zarr support**: Read and write OME-Zarr v0.4, v0.5, and v0.6 (v0.6 adds RFC-5 coordinate systems and transformations)
+- 🗂️ **OME-Zarr support**: Read and write OME-Zarr v0.4, v0.5, and v0.6 (v0.6 adds RFC-5 coordinate systems and transformations), plus the opt-in development version `0.9.dev1` which adopts RFC-3 (extended axis count, names, types and order)
 - 🧪 **Well-tested**: Comprehensive test suite with browser validation
 - 🏗️ **Mirrors Python API**: Familiar interfaces for Python users
 - 📖 **Lazy loading**: Efficient handling of large datasets
@@ -378,7 +378,7 @@ async function fromNgffZarr(
   store: string | MemoryStore | FetchStore,
   options?: {
     validate?: boolean;
-    version?: "0.4" | "0.5" | "0.6";
+    version?: "0.4" | "0.5" | "0.6" | "0.9.dev1";
   }
 ): Promise<NgffMultiscales>
 ```
@@ -414,7 +414,7 @@ async function toNgffZarr(
   store: string,
   multiscales: NgffMultiscales,
   options?: {
-    version?: "0.4" | "0.5" | "0.6";
+    version?: "0.4" | "0.5" | "0.6" | "0.9.dev1";
     chunksPerShard?: number | number[] | Record<string, number>;
   }
 ): Promise<void>
@@ -815,7 +815,7 @@ function upgradeOmeZarr(
   input: string | MemoryStore | FetchStore | Readable,
   options?: {
     output?: string | MemoryStore; // FetchStore is read-only, not a destination
-    version?: "0.4" | "0.5" | "0.6"; // default "0.6"
+    version?: "0.4" | "0.5" | "0.6" | "0.9.dev1"; // default "0.6"
     validate?: boolean;
     overwrite?: boolean; // write-to-new-store only; default true
   },
@@ -870,8 +870,10 @@ await upgradeOmeZarr(store, { version: "0.6" });
 
 **Write-to-new-store.** When `output` is a store distinct from `input`, the
 source is read lazily and re-written to `output` at the requested version
-through the standard write pipeline. Every supported transition (0.4, 0.5, 0.6,
-in either direction) works in this mode, and the source store is never mutated.
+through the standard write pipeline. Every transition among 0.4, 0.5, 0.6 and
+0.9.dev1 works in this mode whenever the target version can express the axis
+model, and the source store is never mutated. An RFC-3 axis model is refused
+below 0.9.dev1, so a store using one only converts upward.
 
 ```typescript
 import { upgradeOmeZarr, type MemoryStore } from "@fideus-labs/ngff-zarr";
--- a/py/ngff_zarr/v04/zarr_metadata.py
+++ b/py/ngff_zarr/v04/zarr_metadata.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
     from ..ngff_image import NgffImage
     from ..v05.zarr_metadata import Metadata as Metadata_v05
     from ..v06.zarr_metadata import Metadata as Metadata_v06
+    from ..v09.zarr_metadata import Metadata as Metadata_v09
 
 logger = logging.getLogger(__name__)
 
@@ -79,6 +80,13 @@ TimeUnits = Union[
     Literal["zettasecond"],
 ]
 Units = Union[SpaceUnits, TimeUnits]
+
+#: Axis unit as RFC-3 allows it: the controlled vocabulary, or any other
+#: string. Every published axes schema declares ``"unit": {"type": "string"}``
+#: with no ``enum``, and an axis of an arbitrary type has no unit in the closed
+#: space/time vocabulary. The union keeps the vocabulary so editors still
+#: complete it. Mirrors the TypeScript port's ``AxisUnit``.
+AxisUnit = Union[Units, str]
 
 supported_dims = ["x", "y", "z", "c", "t"]
 
@@ -163,16 +171,14 @@ def _filter_axis_dict(axis_dict: dict) -> dict:
     Logs a warning if unknown fields are encountered.
 
     Raises:
-        ValueError: If required fields 'name' or 'type' are missing from the axis dictionary.
+        ValueError: If the required field 'name' is missing from the axis
+            dictionary.
     """
-    # Check for required fields before filtering
+    # `name` is the only required axis field: the v0.4 schema accepts an axis
+    # that declares no `type`, so a missing one is read as undefined.
     if "name" not in axis_dict:
         raise ValueError(
             f"Axis dictionary is missing required field 'name': {axis_dict}"
-        )
-    if "type" not in axis_dict:
-        raise ValueError(
-            f"Axis dictionary is missing required field 'type': {axis_dict}"
         )
 
     axis_fields = _get_axis_fields()
@@ -183,13 +189,15 @@ def _filter_axis_dict(axis_dict: dict) -> dict:
             f"Ignoring unknown fields {unknown_fields} in axis '{axis_name}'. "
             f"These fields are not part of the OME-NGFF v0.4 specification."
         )
-    return {k: v for k, v in axis_dict.items() if k in axis_fields}
+    filtered = {k: v for k, v in axis_dict.items() if k in axis_fields}
+    filtered.setdefault("type", None)
+    return filtered
 
 
 @dataclass
 class Axis:
     name: SupportedDims
-    type: AxesType
+    type: AxesType | None
     unit: Units | None = None
     orientation: AnatomicalOrientation | None = None
 
@@ -343,7 +351,7 @@ class Metadata:
 
     def to_version(
         self, version: Union[str, NgffVersion]
-    ) -> Union["Metadata", "Metadata_v05", "Metadata_v06"]:
+    ) -> Union["Metadata", "Metadata_v05", "Metadata_v06", "Metadata_v09"]:
         if isinstance(version, str):
             # raise error for invalid version string
             version = NgffVersion(version)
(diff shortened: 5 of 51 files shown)


py/ngff_zarr/v04/zarr_metadata.py
```diff
@@ -183,13 +179,16 @@ def _filter_axis_dict(axis_dict: dict) -> dict:
...
+    return filtered
 
 
 @dataclass
 class Axis:
-    name: SupportedDims
-    type: AxesType
+    # RFC-3: any axis name, any type; type is optional.
+    name: str
+    type: str | None
```
thewtex: Create a new v1.0/zarr_metadata.py module, include the Axis definition for RFC3 there.

thewtex: Also look at the data structure used in ome-zarr-py

thewtex (changes requested):

thewtex:
> The channel-last order warns rather than raises. ngff-zarr itself writes a
non-spec axis order for vector and RGB data: the channel axis ends up last
(z, y, x, c) via bin-shrink, dask-image and the TIFF S axis, and six tests
exercise it. Raising would break RGB TIFF conversion, so this case warns.
Probably worth its own issue.

@vboussot address this in a separate pr

If review comments in this thread have been addressed, mark them as "Resolved".

py/ngff_zarr/v04/zarr_metadata.py
vboussot: Done — `py/ngff_zarr/v10/zarr_metadata.py` now holds the RFC-3 `Axis` and `Metadata`, with the unchanged RFC-5 pieces re-exported from v06.

ts/src/io/from_ngff_zarr.ts
```diff
@@ -21,7 +21,7 @@ export interface FromOmeZarrOptions {
   /** Enable schema validation of OME-Zarr metadata. */
   validate?: boolean;
   /** Expected OME-Zarr version. */
-  version?: "0.4" | "0.5" | "0.6";
+  version?: "0.4" | "0.5" | "0.6" | "1.0-DEV";
```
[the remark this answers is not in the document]
vboussot: The reader routes detected `1.0-DEV` stores through the v0.6 path (`fromZarrAttrsV06`) — same coordinate-system layout.

ts/src/types/zarr_metadata.ts
```diff
@@ -23,7 +23,8 @@ export interface AxisOrientation {
 
 export interface Axis {
   name: SupportedDims;
-  type: AxesType;
+  // The v0.4 schema accepts an axis that declares no `type`.
+  type: AxesType | undefined;
   unit: Units | undefined;
   orientation?: AxisOrientation | AnatomicalOrientation | undefined;
   discrete?: boolean;
```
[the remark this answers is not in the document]
vboussot: Widened in 23df0f2 — `AxisName`/`AxisType` accept any string while keeping the literal completions.

py/ngff_zarr/v10/zarr_metadata.py
```diff
@@ -0,0 +1,377 @@
...
+            )
+
+        if not all(output == output_cs[0] for output in output_cs):
+            raise ValueError(
+                "Multiple different outputs coordinate systems found in"
+                f" coordinate transformations for multiscales: {output_cs}. "
+                "This is out of spec for this ome-zarr 1.0-DEV."
+            )
+
+        return next(cs for cs in self.coordinateSystems if cs.name == output_cs[0].name)
```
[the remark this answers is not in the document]
vboussot: Fixed in 23df0f2 — explicit lookup with a clear `ValueError`, and `axes`/`dimension_names` now go through `intrinsic_coordinate_system` rather than index 0.

ts/src/utils/structural_validation.ts
```diff
@@ -335,6 +368,13 @@ export function validateSpatialAxisOrder(metadata: Metadata): void {
       "multiscales[0].axes",
     );
```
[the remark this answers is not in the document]
vboussot: Fixed in 23df0f2, on both the TS and Python side so the messages stay in parity.

ts/src/io/to_ngff_zarr_ozx_common.ts
```diff
@@ -16,7 +16,98 @@ import {
...
+  SpecRule,
+  validateAxisCount,
+  validateAxisNamesUnique,
+  validateAxisOrder,
+  validateAxisType,
+  validateSpatialAxisOrder,
+  ValidationError,
+} from "../utils/structural_validation.ts";
+
+/**
```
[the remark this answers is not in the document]
vboussot: Moved onto `gateAxisModel` in 23df0f2.

ts/src/types/zarr_metadata.ts
```diff
@@ -23,7 +23,8 @@ export interface AxisOrientation {
 
 export interface Axis {
   name: SupportedDims;
-  type: AxesType;
+  // The v0.4 schema accepts an axis that declares no `type`.
+  type: AxesType | undefined;
   unit: Units | undefined;
```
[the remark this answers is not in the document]
vboussot: Widened in 23df0f2: `AxisName`/`AxisType` keep the completions but accept any string, the axis rules take `Pick<Metadata, "axes">`, and the `as never` casts are gone.

vboussot:
@thewtex Opened #623 for the channel-last order`to_multiscales` and the pyramidal-TIFF path now normalize axes to (t, c, z, y, x), so the gate's warning only remains for hand-built metadata.

Review threads above are addressed and resolved, ready for another look 👍

py/ngff_zarr/v10/zarr_metadata.py
```diff
@@ -0,0 +1,384 @@
+# SPDX-FileCopyrightText: Copyright (c) Fideus Labs LLC
+# SPDX-License-Identifier: MIT
+"""OME-Zarr ``1.0-DEV`` metadata model (RFC-3).
```
thewtex: We will need to change the package from `v10` to `v09`. The version number string will be initially v0.9.dev1

vboussot: Sure !

vboussot: Will do, `v10` -> `v09`.

On the string itself, which spelling do you want written to disk? Upstream [ome/ngff-spec#190](https://github.com/ome/ngff-spec/pull/190) ("WIP: 0.9dev1", opened today against the new `0.9dev` branch) uses `0.9dev1`, no dot: `schemas/_version.schema` enum becomes `"0.9dev1"` and the `$id`s become `https://ngff.openmicroscopy.org/0.9dev1/schemas/...`. 

1. write `0.9dev1` to match upstream, or
2. write `0.9.dev1` as you wrote it, and accept both spellings on read.

Say which and I will push it.

thewtex: @jni @jo-mueller any guidance on this?

vboussot: Settled upstream, and it is `0.9.dev1` with the dot.

[jni on ome/ngff-spec#190](https://github.com/ome/ngff-spec/pull/190#issuecomment-5349893614):

> my confusion indeed stemmed from the inconsistency between 0.6.devN and 0.6rcN noted by @clbarnes above. I think consistency with the 0.6 series makes sense, which means I'll update this PR to match that. In short @melissalinkert @sbesson, let's go with 0.9.dev1, no need to change your outputs, I will update this PR accordingly.

Applied in ome/ngff-spec@63be7294 "Replace 0.9dev1 with 0.9.dev1": `schemas/_version.schema` now enumerates `0.9.dev1` and every `$id` reads `https://ngff.openmicroscopy.org/0.9.dev1/schemas/`. glencoesoftware/bioformats2raw#330 has written the same string since a562d9fe.

Branch renamed accordingly: `py/ngff_zarr/v10` to `v09`, `NgffVersion.V10DEV` to `V09dev1`, and the version string in both ports.

Worth noting for the description: ome/ngff-spec#190 defines 0.9.dev1 as 0.6rc1 plus RFC-3 **and RFC-4**, not RFC-3 alone. RFC-4 orientation already round-trips at 0.9.dev1 here, since the model reuses the v0.6 axes, so this is a wording fix rather than missing work.

py/test/test_rfc3_axes.py
```diff
@@ -0,0 +1,172 @@
...
+import pytest
+import zarr
+from ngff_zarr import from_ngff_zarr, to_ngff_zarr
+from packaging import version as packaging_version
+
+#: Writing OME-Zarr 0.5 and later needs a Zarr v3 hierarchy.
+needs_zarr_v3 = pytest.mark.skipif(
+    packaging_version.parse(zarr.__version__) < packaging_version.parse("3.0.0b2"),
+    reason="zarr version >= 3.0.0b2 required for OME-Zarr version >= 0.5",
+)
```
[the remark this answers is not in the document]
vboussot: Not applied: both thresholds are established here, and this marker matches the more common one.

`3.0.0b2` with the reason `"zarr version >= 3.0.0b2 required for OME-Zarr version >= 0.5"` appears 14 times across the test suite, and this marker reproduces it verbatim. `3.0.0b1` is used with a different reason, tied to a specific OME-Zarr version ("OME-Zarr v0.5 requires zarr-python >= 3.0.0b1", "OME-Zarr v0.6 requires ..."), which is why `test_v09_metadata.py` uses it for the 0.9.dev1 hierarchy.

The two markers state two different things, so they follow the two existing conventions rather than diverging from one.

thewtex:
> The channel-last order warns rather than raises. ngff-zarr itself writes a
non-spec axis order for vector and RGB data: the channel axis ends up last
(z, y, x, c) via bin-shrink, dask-image and the TIFF S axis, and six tests
exercise it. Raising would break RGB TIFF conversion, so this case warns.
Probably worth its own issue.

I think this was fixed in https://github.com/fideus-labs/ngff-zarr/pull/623

We should strongly prefer write with order (t, c, z, y, x). This should be the default. And we should warn if anything else is used.

docs/python.md
```diff
@@ -213,7 +213,7 @@ To read an OME-Zarr file, use [`from_ngff_zarr`], which returns the
 >>> multiscales = nz.from_ngff_zarr('cthead1.ome.zarr')
 ```
 
-OME-Zarr version 0.1 to 0.6 is supported. Version 0.6 adds RFC-5 coordinate systems and transformations.
+OME-Zarr version 0.1 to 0.6 is supported. Version 0.6 adds RFC-5 coordinate systems and transformations. The opt-in development version `0.9.dev1` additionally adopts RFC-3, lifting every restriction on axis count, names, types and order.
```
thewtex:
```suggestion
OME-Zarr version 0.1 to 0.6 is supported. Version 0.6 adds RFC-5 coordinate systems and transformations. The opt-in development version `0.9.dev1` additionally adopts RFC-3, that expands supported axis counts, names, types and order.
```

docs/spec_features.md
```diff
@@ -35,6 +35,11 @@ supported by `ngff-zarr`.
 - **OME-Zarr v0.4 to v0.6**: Writes OME-Zarr versions 0.4 to 0.6, including
   RFC-4 anatomical orientation. v0.6 additionally adds RFC-5 coordinate systems
   and transformations.
+- **OME-Zarr 0.9.dev1**: Reads and writes the development version that adopts
+  RFC-3, which lifts every restriction on the number, names, types and order
```
thewtex:
```suggestion
  RFC-3, which extends support for the number, names, types and order
```

docs/typescript.md
```diff
@@ -9,7 +9,7 @@ NGFF-Zarr provides a TypeScript implementation for working with OME-Zarr data st
 - 🦕 **Deno-first**: Built for Deno with first-class TypeScript support
 - 📦 **Universal compatibility**: Works in Deno, Node.js, and browsers
 - 🔍 **Type-safe**: Full TypeScript support with Zod schema validation
-- 🗂️ **OME-Zarr support**: Read and write OME-Zarr v0.4, v0.5, and v0.6 (v0.6 adds RFC-5 coordinate systems and transformations)
+- 🗂️ **OME-Zarr support**: Read and write OME-Zarr v0.4, v0.5, and v0.6 (v0.6 adds RFC-5 coordinate systems and transformations), plus the opt-in development version `0.9.dev1` which adopts RFC-3 (unrestricted axis count, names, types and order)
```
thewtex:
```suggestion
- 🗂️ **OME-Zarr support**: Read and write OME-Zarr v0.4, v0.5, and v0.6 (v0.6 adds RFC-5 coordinate systems and transformations), plus the opt-in development version `0.9.dev1` which adopts RFC-3 (extended axis count, names, types and order)
```

py/ngff_zarr/v09/zarr_metadata.py
```diff
@@ -0,0 +1,384 @@
...
+
+
+@dataclass
+class Axis:
+    """An RFC-3 axis.
+
+    Same fields as :class:`ngff_zarr.v06.zarr_metadata.Axis`, with ``name``,
+    ``type`` and ``unit`` unrestricted. The 0.6 axes schema declares
+    ``"unit": {"type": "string"}`` with no ``enum``, and an axis of an
+    arbitrary type has no unit in the closed space/time vocabulary.
```
thewtex: This is a regression. We should continue to use the enum / controlled vocabulary.

vboussot:
@thewtex Thanks, you're right. This PR was based on main from before #623, so it did not carry that normalization. Rebased onto main.

On the axis order: I removed the warning and made it raise like the other MUSTs, so the writer now refuses exactly what `validate_structural` rejects. With #623 normalizing generated axes, the Python path produces canonical
order anyway.

One thing that surfaced: #623 is Python only. The TypeScript port has no canonical-order normalization, and no lazy transpose to build one on, so a non-canonical model that Python normalizes is refused by the TS writer below
0.9.dev1. That looks like its own issue.

py/test/test_rfc3_axes.py
```diff
@@ -0,0 +1,172 @@
...
+@needs_zarr_v3
+@pytest.mark.parametrize("version", ["0.4", "0.5", "0.6", "0.9.dev1"])
+def test_conventional_axes_write_at_every_version(tmp_path, version):
+    """A conventional (z, y, x) image still writes everywhere, 0.9.dev1 included."""
+    axes = [{"name": n, "type": "space"} for n in "zyx"]
+    src = _write_v04(tmp_path / f"zyx-{version}-src.ome.zarr", axes, (2, 3, 4))
+    multiscales = from_ngff_zarr(src, validate=False)
+
+    out = str(tmp_path / f"zyx-{version}-out.ome.zarr")
+    to_ngff_zarr(out, multiscales, version=version)
```
thewtex: We should be using the canonical `from_ome_zarr`, `to_ome_zarr` instead of `from_ngff_zarr`, `to_ngff_zarr`.

README.md
```diff
@@ -67,7 +69,7 @@ The TypeScript package provides universal OME-Zarr support for modern JavaScript
 - 🦕 **Deno-first** with first-class TypeScript support
 - 📦 **Universal compatibility** - Works in Deno, Node.js, and browsers
 - 🔍 **Type-safe** with Zod schema validation
-- 🗂️ **OME-Zarr v0.4, v0.5, and v0.6** support (v0.6 adds RFC-5 coordinate systems and transformations)
+- 🗂️ **OME-Zarr v0.4, v0.5, and v0.6** support (v0.6 adds RFC-5 coordinate systems and transformations), plus the opt-in development version `0.9.dev1` which adopts RFC-3 (unrestricted axis count, names, types and order)
```
thewtex:
```suggestion
- 🗂️ **OME-Zarr v0.4, v0.5, and v0.6** support (v0.6 adds RFC-5 coordinate systems and transformations), plus the opt-in development version `0.9.dev1` which adopts RFC-3 (expanded axis count, names, types and order)
```

thewtex (changes requested):

thewtex:
> One thing that surfaced: #623 is Python only. The TypeScript port has no canonical-order normalization, and no lazy transpose to build one on, so a non-canonical model that Python normalizes is refused by the TS writer below
0.9.dev1. That looks like its own issue.

@vboussot could you please create a PR to address this?

py/ngff_zarr/v09/zarr_metadata.py
```diff
@@ -0,0 +1,389 @@
...
+
+    ``orientation`` (RFC-4) and ``discrete`` (RFC-5) are carried so a 0.6
+    document round-trips without loss.
+    """
+
+    name: AxisName
+    type: AxesType = None
+    unit: str | None = None
```
thewtex: This should be type `Unit | str | None`

vboussot: Done in 7652f15: `unit: AxisUnit | None` with `AxisUnit = Units | str`, so the type reads `Units | str | None` as you asked. Exported from the package root and used on `NgffImage.axes_units`, `to_ngff_image` and the readers. TypeScript mirrors it as `Units | (string & Record<never, never>)` to keep completion on the literal union.

thewtex:
@vboussot please squash this branch into logical commits and force-push

vboussot:
@coderabbitai review

thewtex (approved):

vboussot:
@coderabbitai review

PR #632, opened 2026-08-07
Add itk_transform_resample_bounding_box

### LINKED ISSUE #626: Add itk_transform_resample_bounding_box
Function in Python and TypeScript.

Uses:

  https://github.com/InsightSoftwareConsortium/ITK-Wasm/pull/1549

Apply to perform out-of-core resampling in:

  https://github.com/InsightSoftwareConsortium/ITKElastix/pull/375

vboussot:
Closes #626.

Stacked on #628, which has merged; the base is back on `main`.

Given an ITK transform (including the `CompositeTransform` an Elastix
registration returns) plus a fixed and a moving `NgffImage`, this returns the
region of the moving image a resample would read. Everything comes from
geometry: no pixels are read and no Dask graph is computed.

```python
region = nz.itk_transform_resample_bounding_box(transform, fixed_block, moving)
block = region.crop(moving)   # lazily sliced, translation corrected
```

`crop` is what removes the whole-moving-image materialization in
InsightSoftwareConsortium/ITKElastix#375. It returns `None` when the transformed
grid does not overlap, so a tiling loop can skip that block. Python and
TypeScript.

## Conventions

The result is keyed by dimension name, because the underlying pipeline reports
arrays fastest-axis-first, the reverse of the Zarr order. Start indices may be
negative when the grid extends past the moving origin, so `crop`, `slices` and
`clamped` clamp rather than letting a negative index wrap. The image geometry is
built the way `ngff_image_to_itk_image` builds it, RFC-4 direction included, so
the transform is applied in the space a registration produced it in.

Expected regions in the tests do not come from the pipeline under test: they
come either from the worked examples in the ITK-Wasm documentation or from a
NumPy oracle that recomputes the region from first principles in NGFF axis
order.

## Deformable transforms

Supported: a linear transform is bounded exactly by its transformed grid
corners, while a non-linear one needs the full grid boundary walked, which the
pipeline does.

`itk.BSplineTransform` currently aborts inside `itkwasm-downsample`. That is an
upstream defect, fixed in InsightSoftwareConsortium/ITK-Wasm#1581; displacement
fields and every linear parameterization work today. The limitation is noted in
the docs.

## Summary by CodeRabbit

* **New Features**
  * Added transform-based resampling bounding-box support for Python, browser, Node, and Deno.
  * Added lazy cropping, padding, clamping, selections, and cropped-translation utilities.
  * Added support for preserving Zarr v3 codecs and configuring compressors.
* **Bug Fixes**
  * Improved Gaussian downsampling for integer image types with accurate rounding and range handling.
  * Prevented image and callback inputs from being modified during conversion and writing.
* **Documentation**
  * Added usage guidance for resampling bounding boxes and related TypeScript APIs.


thewtex (commented):
@vboussot nicely done!

Please mark resolve review comments as Resolved.

thewtex (approved):
🥇

PR #644, opened 2026-08-14
feat: model the remaining RFC-5 transforms in both ports

### LINKED ISSUE #562: 0.6: Remaining transforms
Right now, ngff-zarr supports the following transforms:

- Identity
- Scale
- Translate
- Sequence
- Rotation
- Affine

which leaves the following to implement:

- [ ] mapAxis
- [ ] projectAxis
- [ ] byDimension
- [x] displacement and coordinates

Since ngff-zarr won't actually traverse the transform graph and apply transforms, implementing this shouldn't be difficult.

vboussot:
Closes #562.

Completes the RFC-5 / OME-Zarr 0.6 coordinate-transformation model in both ports. As before, ngff-zarr models, reads, writes and validates these transforms; it does not apply them or traverse the transformation graph, so no new dependencies are involved.

## What's added

- **`mapAxis`**: an axis permutation stored as a transpose vector of zero-based integer indices, per the published RFC-5 (the value at position `i` names the input axis that becomes the `i`-th output axis).
- **`byDimension`**: a high dimensional transform built from lower dimensional ones. Each item wraps a transformation with the `input_axes` / `output_axes` (integer indices into the parent's coordinate systems) it applies to.
- **`bijection`**: an explicit forward/inverse transformation pair.

Both ports parse, serialize and round-trip the three types, and enforce the RFC-5 constraints the JSON schema cannot express. This happens at read time, and via `validate_transform` / `validateV06Transform` for programmatically built transforms:

- a `mapAxis` must be a permutation (every zero-based input axis index exactly once), with its length matching resolved input/output coordinate systems;
- `byDimension` items must produce every output axis exactly once, with non-negative indices and axis lists dimensionally consistent with their child transformation; coverage is checked against the output coordinate system when it resolves;
- a `bijection` must join coordinate systems of equal dimensionality.

## Notes

- `projectAxis` is not part of the published RFC-5, so there is nothing to implement for it.
- **TypeScript breaking change (zod)**: `MapAxisTransformationSchema` declared a `Record<string, string>` name-to-name mapping that matches no published spec; it now declares the RFC-5 integer transpose vector. The `byDimension` schema gains the wrapped item structure for the same reason.
- The 0.6 JSON schema already covered all three types; the `byDimension` axis arrays tighten from `number` to non-negative `integer`.
- Follow-up candidates, out of scope here: the TS zod schemas declare an `inverseOf` type absent from the published RFC-5, and their `rotation` payload is a flat array where the rest of the code uses a 2D matrix.

## Testing

- py: write→read round-trips for all three types (by value), plus seven rejection cases (non-permutation `mapAxis`, duplicate/missing output axes, dimension mismatches); related suites (convert, structural validation, rfc9, from_ngff_zarr) pass.
- ts: same round-trips and rejections; the full suite (490 tests) passes.

## Summary by CodeRabbit

- **New Features**
  - Added RFC 5 coordinate transformations for axis mapping, dimension-specific transformations, and reversible transformation pairs.
  - Added recursive support for nested transformations in v0.6 metadata.
  - Added serialization, parsing, and construction support across Python and TypeScript.

- **Validation**
  - Added checks for axis permutations, dimensionality, coverage, integer indices, and compatible reversible transformations.
  - Added validation for nested transformations and coordinate-system references.

- **Documentation**
  - Expanded RFC and schema documentation with the new transformation types and constraints.

### DIFF
--- a/py/ngff_zarr/v06/zarr_metadata.py
+++ b/py/ngff_zarr/v06/zarr_metadata.py
@@ -211,6 +211,13 @@ def _item_dimensions(transformation: Transform) -> int | None:
     return None
 
 
+def _require_integer_axes(axes: list, context: str) -> None:
+    """Axis indices are zero-based integer positions; reject anything else."""
+    for axis in axes:
+        if isinstance(axis, bool) or not isinstance(axis, int):
+            raise ValueError(f"{context} axis indices must be integers; got {axis!r}")
+
+
 def validate_transform(
     transformation: Transform,
     coordinateSystems: list[CoordinateSystem] | None = None,
@@ -229,6 +236,7 @@ def validate_transform(
 
     if isinstance(transformation, MapAxis):
         indices = transformation.mapAxis
+        _require_integer_axes(indices, "mapAxis")
         if sorted(indices) != list(range(len(indices))):
             raise ValueError(
                 "mapAxis must be a permutation holding every zero-based "
@@ -242,12 +250,22 @@ def validate_transform(
                     f"{count} axes of coordinate system '{identifier.name}'"
                 )
     elif isinstance(transformation, ByDimension):
+        input_count = _axis_count(transformation.input, coordinateSystems)
         seen_output_axes: set[int] = set()
         for item in transformation.transformations:
             axes = list(item.input_axes) + list(item.output_axes)
+            _require_integer_axes(axes, "byDimension")
             if any(axis < 0 for axis in axes):
                 raise ValueError(
                     f"byDimension axis indices must be non-negative; got {axes}"
+                )
+            if input_count is not None and any(
+                axis >= input_count for axis in item.input_axes
+            ):
+                raise ValueError(
+                    f"byDimension input axes {item.input_axes} exceed the "
+                    f"{input_count} axes of coordinate system "
+                    f"'{transformation.input.name}'"
                 )
             duplicated = seen_output_axes.intersection(item.output_axes)
             if duplicated or len(set(item.output_axes)) != len(item.output_axes):
@@ -744,6 +762,11 @@ class Metadata:
 
             transformation.input = input
             transformation.output = output
+            # The wrapper branches above construct their transform from the
+            # payload fields alone, so the optional name is restored here for
+            # every type.
+            if transform.get("name") is not None:
+                transformation.name = transform["name"]
             validate_transform(transformation, coordinateSystems)
             parsed_transforms.append(transformation)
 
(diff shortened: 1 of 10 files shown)


py/ngff_zarr/v06/zarr_metadata.py
```diff
@@ -127,17 +139,178 @@ class Displacements(BaseTransform):
...
+    if isinstance(transformation, MapAxis):
+        return len(transformation.mapAxis)
+    return None
+
+
+def _require_integer_axes(axes: list, context: str) -> None:
+    """Axis indices are zero-based integer positions; reject anything else."""
+    for axis in axes:
+        if isinstance(axis, bool) or not isinstance(axis, int):
+            raise ValueError(f"{context} axis indices must be integers; got {axis!r}")
```
jo-mueller: Maybe these could be class methods of the transform classes? Any transform class could then expose a `transform_instance.axis_count()` function that will return simply None, if input/output aren't specified?

py/ngff_zarr/v06/zarr_metadata.py
```diff
@@ -127,17 +139,178 @@ class Displacements(BaseTransform):
...
+    spec: a ``mapAxis`` that is not a permutation, a ``byDimension`` whose
+    items skip or duplicate an output axis, or a ``bijection`` between
+    coordinate systems of different dimensionality. Dimension checks against
+    the ``input`` and ``output`` coordinate systems run when those resolve
+    to a known system and are skipped otherwise, as for the wrapped
+    transforms that omit them.
+    """
+    coordinateSystems = coordinateSystems or []
+
+    if isinstance(transformation, MapAxis):
```
jo-mueller: Similarly to my previous transform, I would find it cleaner to have this validation done in a post-hoc init of the transforms dataclasses 

That way, every transfrom's validation happens closer to where the dataclass is actually defined

py/ngff_zarr/v06/zarr_metadata.py
```diff
@@ -536,6 +709,30 @@ def _parse_transforms(
...
+                        [item["transformation"]], coordinateSystems
+                    )
+                    items.append(
+                        ByDimensionItem(
+                            transformation=sub_transform,
+                            input_axes=list(item["input_axes"]),
+                            output_axes=list(item["output_axes"]),
+                        )
+                    )
+                transformation = ByDimension(transformations=items)
```
jo-mueller: Maybe this could also be "hidden" in the `ByDimension`'s init or post-init?

thewtex (approved):
👍

PR #659, opened 2026-08-20
fix(py): resolve cross-file JSON Schema references during validation

### LINKED ISSUE #647: JSON schema validation cannot resolve the 0.6 coordinate_transformations schema reference
`ngff_zarr.validate.validate(attrs, version="0.6")` fails on any 0.6 image because the 0.6 `image.schema` references the separate `coordinate_transformations.schema` by its `https://ngff.openmicroscopy.org/0.6.dev4/...` URL, and the validator does not map that URL to the bundled local file.

```python
import json
import tempfile
import numpy as np
import ngff_zarr as nz
from ngff_zarr.validate import validate

image = nz.to_ngff_image(np.zeros((4, 8, 8), dtype=np.float32), dims=["z", "y", "x"])
multiscales = nz.to_multiscales(image, scale_factors=[2])
with tempfile.TemporaryDirectory() as path:
    nz.to_ngff_zarr(path, multiscales, version="0.6")
    attrs = json.load(open(path + "/zarr.json"))["attributes"]
    validate(attrs, version="0.6")
```

```
jsonschema.exceptions._WrappedReferencingError: Unresolvable:
https://ngff.openmicroscopy.org/0.6.dev4/schemas/coordinate_transformations.schema#/$defs/scale
```

The same call passes for 0.4 and 0.5, whose schemas are self-contained. Registering the bundled 0.6 schema files in a `referencing.Registry` keyed by their `$id` resolves the reference; validated locally, the bundled 0.6 schemas then accept what the 0.6 writer produces.

Reproduced on `main` at 55ab973.

vboussot:
`validate(attrs, version="0.6")` failed on any 0.6 image:

```
jsonschema.exceptions._WrappedReferencingError: Unresolvable:
https://ngff.openmicroscopy.org/0.6.dev4/schemas/coordinate_transformations.schema#/$defs/scale
```

From 0.6 the spec splits axes, coordinate systems and coordinate transformations into separate files that `image.schema` reaches by absolute `$id` URL. Validation is offline, so nothing dereferences those URLs, and only the requested schema (plus the `_version` one from 0.5) was in the registry.

## The fix

Build the `referencing.Registry` from every bundled schema for the version, each keyed by its own `$id`. Every non-internal `$ref` in every bundled schema points at a sibling file in the same version directory, so a per-version registry is complete.

All 52 bundled schemas across 0.1 to 0.6 declare an `$id`, and no `$id` is repeated within a version, so the registry is unambiguous.

A pre-release resolves to the tree of the release it leads to, so the `"0.6.dev4"` string a 0.6 store records on disk maps to `spec/0.6`. The version is matched against the bundled directory names rather than joined onto the path, because it arrives straight from a store's metadata whenever the caller passes an explicit version. An unbundled version now names the bundled ones instead of raising `FileNotFoundError` on an internal path.

## Two widened behaviours

**`strict=True` works at every version.** The `strict_*` schemas wrap their base schema by absolute `$id`, and the pre-0.6 ones omit `$schema`, so they need both the registry and an explicit default specification.

**The v0.6 read path validates.** `from_ngff_zarr(..., validate=True)` checks a 0.6 store. The reader takes the version from the `ome` namespace, where 0.6 records it, falling back to the multiscales entry as the v0.4 path does.

The upstream `0.6rc0` schemas carry the same absolute cross-file references, so bundling them instead would need the same registry.

## Verification

Before and after, on a 0.6 store written by the library:

| | `strict=False` | `strict=True` |
| --- | --- | --- |
| before | `Unresolvable: .../coordinate_transformations.schema` | `Unresolvable: .../image.schema` |
| after | ok | ok |

Closes #647.

## Note for the merge order

This conflicts with #611 in `py/ngff_zarr/validate.py`, and only there. #611 adds a guard that reports 0.9.dev1 as having no published schema; this PR rewrites `load_schema` around a version-to-directory lookup that also rejects unbundled versions. Keeping the #611 guard first preserves its message, which points at `validate_structural`; `test_no_bundled_schema_is_reported_explicitly` expects `NotImplementedError`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected metadata validation so caller-specified settings are respected.
  * Improved schema version detection, including fallback and prerelease handling.
  * Enabled reliable offline validation across bundled schemas and cross-schema references.
  * Added stricter rejection of invalid, unsupported, and forged metadata versions.
  * Improved validation when reading metadata from storage.
  * Strengthened image metadata validation, including strict schema checks.
  * Added clearer errors when schemas are unavailable.

### DIFF
--- a/mcp/ngff_zarr_mcp/tools.py
+++ b/mcp/ngff_zarr_mcp/tools.py
@@ -4,7 +4,6 @@
 
 import tempfile
 from pathlib import Path
-from typing import Any
 
 import zarr
 from ngff_zarr import (  # type: ignore[import-untyped]
@@ -27,7 +26,7 @@ except ImportError:
         version: str = "0.4",
         model: str = "image",
         strict: bool = False,
-    ) -> Any:
+    ) -> None:
         pass
 
 
--- a/py/ngff_zarr/from_ngff_zarr.py
+++ b/py/ngff_zarr/from_ngff_zarr.py
@@ -349,9 +349,8 @@ def from_ome_zarr(
     if version.startswith("0.6"):
         from .v06.zarr_metadata import Metadata
 
-        # TODO: Restore validation for v0.6
         metadata_obj, images = Metadata._from_zarr_attrs(
-            root_attrs, store, validate=False, subpath=subpath
+            root_attrs, store, validate=validate, subpath=subpath
         )
         method, method_type, method_metadata = _extract_method_metadata(
             root_attrs["ome"]["multiscales"][0]
--- a/py/ngff_zarr/v06/zarr_metadata.py
+++ b/py/ngff_zarr/v06/zarr_metadata.py
@@ -681,7 +681,8 @@ class Metadata:
             # per-entry value is the fallback, as on the v0.4 read path.
             schema_version = str(
                 root_attrs["ome"].get("version")
-                or root_attrs["ome"]["multiscales"][0].get("version", "0.6")
+                or root_attrs["ome"]["multiscales"][0].get("version")
+                or "0.6"
             )
             validate_ngff(root_attrs, version=schema_version)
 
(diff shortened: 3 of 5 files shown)


thewtex (approved):
👍

PR #660, opened 2026-08-20
fix(py): validate the coordinate-systems metadata model

### LINKED ISSUE #646: validate_structural raises AttributeError on OME-Zarr 0.6 metadata
`validate_structural` reads `metadata.axes`, which the OME-Zarr 0.6 `Metadata` does not have: in 0.6 the axes live in `metadata.coordinateSystems[0].axes`. Calling it on a 0.6 image raises `AttributeError` before any rule runs.

```python
import tempfile
import numpy as np
import ngff_zarr as nz
from ngff_zarr import validate_structural

image = nz.to_ngff_image(np.zeros((4, 8, 8), dtype=np.float32), dims=["z", "y", "x"])
multiscales = nz.to_multiscales(image, scale_factors=[2])
with tempfile.TemporaryDirectory() as path:
    nz.to_ngff_zarr(path, multiscales, version="0.6")
    validate_structural(nz.from_ngff_zarr(path).metadata)
```

```
File "ngff_zarr/structural_validation.py", line 217, in validate_axis_count
    count = len(metadata.axes)
AttributeError: 'Metadata' object has no attribute 'axes'
```

The same call passes for the 0.4 and 0.5 writers, and `from_ngff_zarr(path, validate=True)` reads the 0.6 image without error since it takes a different path.

Reproduced on `main` at 55ab973.

vboussot:
`validate_structural` reads `metadata.axes`, which the v0.6 `Metadata` does not carry: from 0.6 the axes live in `coordinateSystems`. Calling it on a 0.6 image raised `AttributeError` before any rule ran.

This affects every version in practice, not just stores written at 0.6, because `from_ome_zarr` and `to_multiscales` return the v0.6 model whatever the store holds.

## The fix

v0.6 `Metadata` exposes the intrinsic coordinate system's axes as a property, so the axis rules read them directly.

The dataset rules read a view of the RFC-5 model in the flat v0.4 shape they are written against. From 0.6 each dataset carries a single transform mapping its array to the intrinsic system: an `identity`, a `scale`, or a `sequence` of a `scale` and a `translation`.

The view renders that transform, it does not repair it. An `identity` is synthesized into the unit scale and zero translation it stands for; every other form is rendered in source order, one entry per `scale` or `translation` found, so a dataset carrying no scale, two scales, or a translation ahead of its scale still reaches `validate_per_dataset_scale_count` and `validate_transform_order`. Vectors are taken verbatim, never padded or truncated, so a wrong-length one still reaches `validate_scale_length`.

The TypeScript port arrives at its structural pass from a different starting point: its v0.6 reader replaces each dataset's transform with the single `[scale, translation]` pair `extractScaleTranslation` composes, so the per-dataset shape rules cannot fire there. This port is the stricter of the two at 0.6. The parity contract (rule identifiers, levels, evaluation order) is unchanged.

## Three limits

**Multiscale-level transforms are left out of the view.** From 0.6 they map between *named* coordinate systems whose axis counts need not match the intrinsic system's, so the global arm of `scale-length-mismatch` does not measure those vectors.

**`zarr-format` and `ome-namespace` are inert on the 0.6 model.** Both gate on a `version` field only the v0.4 model carries; from 0.5 on the version lives in the group-level `ome` namespace. Reading it is defensive here, so a model without the field leaves both rules inert instead of raising.

**The axis rules read the intrinsic coordinate system and only it.** A v0.6 document may declare further coordinate systems, and `coordinate_systems.schema` applies `axes.schema` to each of them; those are not validated here.

The rule reference and the validation overview record all three.

## Not in scope

`from_ome_zarr(store, validate=True)` still does not run the structural rules on a 0.6 store: the v0.6 read path validates the schema and RFC 4 only, where the v0.4 and v0.5 paths call `validate_structural`. This PR is the prerequisite for wiring it in; doing so is a separate change, tracked in #667 along with the other validation gaps.

## Verification

Reproduced on a 0.6 store written by the library:

| | |
| --- | --- |
| before | `AttributeError: 'Metadata' object has no attribute 'axes'` |
| after | `validate_structural: OK` |

Full Python suite: 963 passed, 3 skipped. `prek run --all-files` passes.

Closes #646.

## Note for the merge order

This conflicts with #611 in `py/ngff_zarr/structural_validation.py` and `docs/validation/rule-reference.md`, and only there. #611 inserts `axis-names-unique` as rule 4, which shifts the row this PR edits, and adds a `version` parameter to the axis rules; that parameter is passed by the caller, not read off the model, so the view does not interact with it. Merging #611 first keeps both conflicts mechanical.

The vendored 0.6 schema bump to `0.6rc0` does not touch either file, so it is independent of this one in both directions.

## Summary by CodeRabbit

* **New Features**
  * Structural validation now supports OME-Zarr v0.6 coordinate-system metadata alongside v0.4 and v0.5 metadata.
  * Added access to intrinsic coordinate-system axes for v0.6 metadata.

* **Bug Fixes**
  * Improved validation of dataset and coordinate-system transforms, including identity, scale, and sequence transformations.
  * Improved scale-length, dataset-order, axis-count, and orientation validation.

* **Documentation**
  * Updated validation guidance, API examples, and rule references for v0.6 metadata and transform handling.

* **Tests**
  * Added comprehensive coverage for v0.6 validation and OME-Zarr version round trips.


py/test/test_structural_validation_v06.py
```diff
@@ -0,0 +1,235 @@
...
+    CoordinateSystem,
+    CoordinateSystemIdentifier,
+    Dataset,
+    Identity,
+    Metadata,
+    Scale,
+    TransformSequence,
+    Translation,
+)
+from packaging import version as packaging_version
```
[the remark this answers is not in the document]
vboussot: Not applied: `ruff` reverts this exact edit. `py/pyproject.toml` sets `src = ["ngff_zarr"]`, which does not make `ngff_zarr` first-party, so it sorts alphabetically among the third-party imports and `packaging` follows it. Checked with `ruff check --select I --diff`, which proposes moving the line back down.

docs/validation/api.md
```diff
@@ -41,8 +41,10 @@ from ngff_zarr import (
 )
 ```
 
-`validate_structural(metadata, options=None)` runs the image/multiscales rules.
-When `options` is `None` it uses `ValidateOptions()`, i.e.
+`validate_structural(metadata, options=None)` runs the image/multiscales rules
+against either metadata model — the flat v0.4/v0.5 `Metadata` or the v0.6
+`coordinateSystems` one that `from_ngff_zarr` and `to_multiscales` return (see
```
thewtex: from_ome_zarr

docs/validation/overview.md
```diff
@@ -34,6 +34,28 @@ rule has a stable, lower-kebab-case identifier (a [[rule-reference|SpecRule]])
...
+are written against. v0.6 (RFC 5) instead keeps its axes in
+`coordinateSystems` and gives each dataset a single transform — an `identity`,
+a `scale`, or a `sequence` of a `scale` and a `translation` — mapping its array
+to the intrinsic coordinate system; that model is rendered in the flat shape
+before the rules run, so both validate identically. Only an `identity` is
+synthesized, into the unit scale and zero translation it stands for; every
+other form is rendered in source order, one entry per `scale` or `translation`
+found, so a dataset carrying no scale, two scales, or a translation ahead of
+its scale is still rejected. The reduction applies to
+every version in practice, because `from_ngff_zarr` and `to_multiscales` return
```
thewtex: from_ome_zarr

py/ngff_zarr/structural_validation.py
```diff
@@ -10,9 +10,13 @@
...
 
-The rules are pure Python (standard library only) and operate on the already
-parsed :class:`~ngff_zarr.v04.zarr_metadata.Metadata` dataclass, so they work
-even when the optional ``[validate]`` extra (``jsonschema``) is not installed.
+The rules are pure Python (standard library only) and operate on already
+parsed metadata, so they work even when the optional ``[validate]`` extra
+(``jsonschema``) is not installed. Both metadata models are accepted: the flat
+:class:`~ngff_zarr.v04.zarr_metadata.Metadata` of v0.4/v0.5 directly, and the
+``coordinateSystems`` model of v0.6 through the reduction in
+:func:`_flat_model` -- which matters because ``from_ngff_zarr`` and
```
thewtex: from_ome_zarr

py/test/test_structural_validation_v06.py
```diff
@@ -0,0 +1,289 @@
...
+        pytest.param("0.6", marks=needs_zarr_v3),
+    ],
+)
+def test_validate_structural_after_round_trip(tmp_path, version):
+    # Every version reads back as the v0.6 model, so this is the same code path
+    # each time; before the reduction it raised AttributeError.
+    image = to_ngff_image(np.zeros((4, 8, 8), dtype=np.float32), dims=["z", "y", "x"])
+    multiscales = to_multiscales(image, scale_factors=[2])
+    store = tmp_path / f"image-{version}.ome.zarr"
+    to_ngff_zarr(str(store), multiscales, version=version)
```
thewtex: to_ome_zarr and throughout this test file

py/test/test_structural_validation_v06.py
```diff
@@ -0,0 +1,289 @@
...
+
+    metadata = from_ngff_zarr(str(store)).metadata
```
thewtex: from_ome_zarr, throughout this test file

thewtex (changes requested):
Thanks @vboussot !

A few minor naming changes throughout this test file.

thewtex (approved):

thewtex (approved):

PR #662, opened 2026-08-20
test(ts): name two test files so Deno collects them

vboussot:
`deno test` collects `*_test.ts`, `*.test.ts` and `test.ts`.
`test/schemas/test_ome_zarr.ts` and `test/test_data_layout.ts` match none of
those and nothing imports them, so their 23 cases have never run: the Zod
schemas for the OME-Zarr, image, label, plate and well documents, and the ITK
to NGFF data-layout round trip.

Renamed to the collected form. All 23 pass, so nothing was hiding behind the
naming; the suite goes from 552 to 575.


thewtex (approved):
💯

PR #663, opened 2026-08-20
refactor(py,ts): keep the RFC-4 vocabulary in one place per port

vboussot:
The 24 anatomical orientation values were transcribed by hand four times per
language, none derived from another and none checked against another. Two of
those copies were traps rather than duplication:

- **Python.** `validate_rfc4_orientation` checked membership against its own
  24-value set, then looked the value up in `_ANATOMICAL_AXIS_OF` unguarded. A
  value in the set but not in the table would raise `KeyError` instead of a
  validation error, and the message-marker mapping in `validate_structural`
  would never see it. Membership is now checked against the table, as
  `rfc4_conformance` already did.
- **TypeScript.** `VALID_ORIENTATION_VALUES` stopped validating anything when
  the check moved to the Zod schema; it only formats the "valid values are"
  message, so it could advertise a vocabulary different from the enforced one.
  It is now derived from the pair table.

The copies that remain (the public enum, the bundled schema, the pair table)
are pinned by new mirrored tests: same 24 values, and exactly two directions
per anatomical axis.

Python: 913 passed, 3 skipped. TypeScript: 556 passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RFC 4 anatomical-orientation validation by deriving accepted values from a single authoritative vocabulary.
  * Invalid orientation messages now consistently list the supported values.
  * Non-string and unhashable orientation inputs now return a clear validation error instead of causing unexpected type errors.

* **Tests**
  * Added consistency checks across public vocabulary definitions and the bundled schema.
  * Verified the expected 24 orientation values and two directions for each of 12 anatomical axes.

### DIFF
--- a/py/ngff_zarr/rfc4_validation.py
+++ b/py/ngff_zarr/rfc4_validation.py
@@ -25,7 +25,13 @@ def load_rfc4_orientation_schema() -> dict:
 
 # Each RFC 4 value maps to a stable key for the anatomical axis it lies on, so the
 # two antonyms of a pair (e.g. "left-to-right" / "right-to-left") collapse to one
-# key. Used to enforce "one direction per anatomical axis" (mutual exclusion).
+# key. Used to enforce "one direction per anatomical axis" (mutual exclusion), and
+# its keys are this module's vocabulary: membership is checked against them, and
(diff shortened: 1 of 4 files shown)


thewtex (approved):
🥇

PR #711, opened 2026-09-01
Return units for axes where translation axes present

folterj:
If translation axis present with matching unit, which was not present for scale, the units for those axis still needs to be returned. This is fixes in this PR including more extensive test

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TIFF-to-NGFF conversion by consistently recording translation units for image dimensions.
  * Preserved available OME metadata when pixel-size units are missing or only partially provided.
  * Applied the correct unit for each dimension when converting translation values.

* **Tests**
  * Expanded coverage for complete, missing, and partial unit metadata.
  * Improved test file handling to ensure generated files are cleaned up automatically.


thewtex (approved):
@folterj thank you!

PR #726, opened 2026-09-02
Make the version-scoped metadata API public per version (closes #715)

### LINKED ISSUE #715: Export the RFC-5 metadata dataclasses at the package top level
The RFC-5 metadata dataclasses are only importable from an internal path. `Axis` is exported at the package top level, but `CoordinateSystem`, `CoordinateSystemIdentifier` and `Displacements` must come from `ngff_zarr.v06.zarr_metadata` (checked at 0.45.0).

Code that builds RFC-5 coordinate systems and displacement declarations (the write side of what #708 covers) therefore imports an internal module that any refactor can move without a deprecation cycle, and it breaks at import time when that happens.

Exporting the three names beside `Axis` in `__init__` would settle it. One-line change; happy to send the PR.

vboussot:
Issue #715 asked to export the RFC-5 metadata dataclasses at the package top level. Triaging it surfaced why that cannot be done unambiguously: `CoordinateSystem` is defined in both `v06.zarr_metadata` and `v09.zarr_metadata` with different fields, so a bare `ngff_zarr.CoordinateSystem` would silently pick a spec version and change meaning when the default moves.

So this makes the version-scoped path the public surface instead:

- each `vNN.zarr_metadata` declares `__all__` (the explicit public class list per version);
- the package exposes `v04`, `v06`, `v09` so `ngff_zarr.v06.zarr_metadata` is reachable without a separate import;
- `docs/rfc5.md` says to import from the version you target, and notes that the field transforms (Displacements, Coordinates) are shared and v09 reuses them from v06;
- a test pins the public names and that the two `CoordinateSystem` are distinct classes.

No top-level `CoordinateSystem` export, deliberately: the ambiguity is real and naming it would hide it.

Closes #715

## Summary by CodeRabbit

* **API Changes**
  * Public package access remains available for the released v0.4 and v0.6 metadata APIs.
  * The v0.9 metadata API remains importable through its version-specific path but is no longer re-exported at the package level.
  * Clarified version-specific coordinate systems and shared field transforms.

* **Documentation**
  * Clarified version-scoped metadata usage and API differences.
  * Improved formatting of the codec-chain migration example.

* **Tests**
  * Added coverage for versioned module access, public exports, coordinate-system models, and shared transform models.

### DIFF
--- a/py/ngff_zarr/__init__.py
+++ b/py/ngff_zarr/__init__.py
@@ -3,10 +3,12 @@
 #
 # SPDX-License-Identifier: MIT
 
-# The version-scoped metadata models are public under their own path. RFC-5's
-# CoordinateSystem exists in both v06 and v09 with different fields, so there is no
-# unambiguous top-level export: import from the version you target.
-from . import v04, v06, v09
+# The released metadata models are public under their own path. RFC-5's
+# CoordinateSystem exists in more than one of them with different fields, so
+# there is no unambiguous top-level export: import from the version you target.
+# The 0.9 development model is reachable as ngff_zarr.v09.zarr_metadata and is
+# deliberately not exported here, since it changes between releases.
+from . import v04, v06
 from .__about__ import __version__
 from ._supported_versions import (
     SUPPORTED_VERSIONS,
@@ -134,7 +136,6 @@ from .validate import validate
 __all__ = [
     "v04",
     "v06",
-    "v09",
     "__version__",
     "SUPPORTED_VERSIONS",
     "V06_ONDISK_VERSION",
--- a/py/ngff_zarr/__init__.py
+++ b/py/ngff_zarr/__init__.py
@@ -3,6 +3,10 @@
 #
 # SPDX-License-Identifier: MIT
 
+# The version-scoped metadata models are public under their own path. RFC-5's
+# CoordinateSystem exists in both v06 and v09 with different fields, so there is no
+# unambiguous top-level export: import from the version you target.
+from . import v04, v06, v09
 from .__about__ import __version__
 from ._supported_versions import (
     SUPPORTED_VERSIONS,
@@ -128,6 +132,9 @@ from .v04.zarr_metadata import (
 from .validate import validate
 
 __all__ = [
+    "v04",
+    "v06",
+    "v09",
     "__version__",
     "SUPPORTED_VERSIONS",
     "V06_ONDISK_VERSION",
--- a/py/ngff_zarr/v06/zarr_metadata.py
+++ b/py/ngff_zarr/v06/zarr_metadata.py
@@ -1075,3 +1075,27 @@ class Metadata:
     @property
     def dimension_names(self) -> tuple:
         return tuple([ax.name for ax in self.coordinateSystems[0].axes])
+
+
+#: The public, version-scoped API of this module. CoordinateSystem here is the 0.6 model; ngff_zarr.v09.zarr_metadata carries a different one.
+__all__ = [
+    "Axis",
+    "CoordinateSystem",
+    "CoordinateSystemIdentifier",
+    "BaseTransform",
+    "Identity",
+    "Scale",
+    "Translation",
+    "Rotation",
+    "Affine",
+    "Coordinates",
+    "Displacements",
+    "MapAxis",
+    "ProjectAxis",
+    "ByDimensionItem",
+    "ByDimension",
+    "Bijection",
+    "TransformSequence",
+    "Dataset",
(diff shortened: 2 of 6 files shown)


py/ngff_zarr/v09/zarr_metadata.py
```diff
@@ -397,3 +397,13 @@ def _from_zarr_attrs(
             patched, store, validate=False, subpath=subpath
         )
         return cls._from_v06(metadata_v06), images
+
+
+#: The public, version-scoped API of this module. Axis/CoordinateSystem/Metadata are 0.9's own; the field transforms are reused from v06.
+__all__ = [
```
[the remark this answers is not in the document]
vboussot: Fixed: v09 __all__ now lists its full public surface including the v06 re-exports (Dataset, Scale, Coordinates, Displacements, ...), so import * keeps them.

py/test/test_versioned_metadata_api.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+
+import ngff_zarr
+from ngff_zarr.v06 import zarr_metadata as v06
+from ngff_zarr.v09 import zarr_metadata as v09
+
+
+def test_version_subpackages_are_reachable_from_the_package():
+    assert ngff_zarr.v06.zarr_metadata.CoordinateSystem is v06.CoordinateSystem
+    assert ngff_zarr.v09.zarr_metadata.CoordinateSystem is v09.CoordinateSystem
```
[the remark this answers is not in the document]
vboussot: Fixed: the test now asserts the v04 path, the top-level {v04, v06, v09} exports, and the required version-specific names (CoordinateSystemIdentifier in v06; Coordinates/Displacements/Dataset/TransformSequence in v09).

py/test/test_versioned_metadata_api.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_each_module_declares_its_public_surface():
+    for name in v06.__all__:
+        assert hasattr(v06, name), f"v06.__all__ names {name}, which is absent"
+    for name in v09.__all__:
+        assert hasattr(v09, name), f"v09.__all__ names {name}, which is absent"
+    assert "CoordinateSystem" in v06.__all__ and "CoordinateSystem" in v09.__all__
```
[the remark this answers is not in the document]
vboussot: Fixed: the compound assertion is split into two, and ruff check passes.

py/ngff_zarr/v06/zarr_metadata.py
```diff
@@ -1075,3 +1075,27 @@ def _parse_transforms(
     @property
     def dimension_names(self) -> tuple:
         return tuple([ax.name for ax in self.coordinateSystems[0].axes])
+
+
+#: The public, version-scoped API of this module. CoordinateSystem here is the 0.6 model; ngff_zarr.v09.zarr_metadata carries a different one.
+__all__ = [
```
thewtex: We usually put `__all__` at the top of the file

py/ngff_zarr/v09/zarr_metadata.py
```diff
@@ -397,3 +397,26 @@ def _from_zarr_attrs(
             patched, store, validate=False, subpath=subpath
         )
         return cls._from_v06(metadata_v06), images
+
+
+#: The public, version-scoped API of this module: 0.9's own Axis/CoordinateSystem/Metadata,
+#: plus the field transforms and shared types it reuses from v06.
+__all__ = [
```
thewtex: We usually put __all__ at the top of the file

py/ngff_zarr/__init__.py
```diff
@@ -128,6 +132,9 @@
 from .validate import validate
 
 __all__ = [
+    "v04",
+    "v06",
+    "v09",
```
thewtex: v09 is strictly a dev version that is expected to change between releases -- it should not be exposed.

py/ngff_zarr/__init__.py
```diff
@@ -3,6 +3,10 @@
 #
 # SPDX-License-Identifier: MIT
 
+# The version-scoped metadata models are public under their own path. RFC-5's
+# CoordinateSystem exists in both v06 and v09 with different fields, so there is no
+# unambiguous top-level export: import from the version you target.
+from . import v04, v06, v09
```
thewtex: v09 is strictly a dev version that is expected to change between releases -- it should not be exposed.

thewtex (changes requested):
A few changes requested inline

thewtex (approved):
💯

PR #729, opened 2026-09-02
Carry a displacement range through a non-linear outer stage

### LINKED ISSUE #727: More tests for resample_bounding_box
Add tests for transformations:

1. B-spline with non-trivial coefficients
2. Composite b-spline, affine
3. Composite with multiple affine's
4. Composite affine + b-spline with non-trivial coefficients
5. Composite affine + displacement field
6. Composite affine + b-spline + displacement field
7. Composite affine + displacement field 1 + displacement field 2

Verify with `itk.TransformToDisplacementFieldFilter`, putting the result into the same low, high interval bounds functions we use for displacement fields.

vboussot:
Closes #727.

The seven compositions that issue lists are measured against `itk.TransformToDisplacementFieldFilter`: it collapses the whole composition into one displacement field over the fixed grid, so a stage acting strictly inside the grid, which the boundary walk cannot see, shows up in the field. Adding each grid point's own position back gives the positions the transform actually reaches, and the region has to contain them.

Two of the seven did not: an affine over a B-spline over a displacement field, and an affine over two displacement fields, each leaving about 19 pixels outside the reported region.

## What was wrong

#704 bounds a displacement stage by the range its values take, and carries that range through the stages applied after it, one stage at a time. The carry handled an affine, through the signs of its matrix, and the orthonormal family, into the ball its Euclidean reach bounds; anything else gave up and the boundary walk stood. A second displacement stage is exactly that "anything else", so a list holding two of them, or a field under a B-spline, fell back to a walk that reports the identity region for an interior bump.

## What changed

The per-stage fold becomes one accumulation running outward from the last entry, the one ITK applies first. A displacement stage adds its own range, since the list being widened replaced that stage by the identity; an affine maps the accumulated range through the signs of its matrix; an orthonormal stage bounds it by its Euclidean reach. Stages therefore compose, in any nesting, and the fallback now only covers parameterizations that carry no range at all, the velocity fields among them, which the docstring already names.

This is smaller than what it replaces: one pass instead of a fold per stage plus a sum, with no cross-stage bookkeeping. It reports the same bound on every list the fold already handled, which the tests #704 added pin unchanged, and the entries are scanned once for their intervals rather than once per displacement stage, keeping the property that a field's parameters are read once.

## Tests

The seven cases are covered in Python against the filter oracle. Case 3, several affines, is purely linear, so the region is asserted equal to the reached extent rather than merely containing it: the walk reports a linear map exactly and nothing should widen it.

One further test does what the issue describes literally: it reads the materialized field with `_stage_interval`, the same function a field stage goes through, and checks the range the split derives holds it. That comparison only means something when the linear part is the identity, so the composition it uses carries no affine, and the test asserts that before comparing.

TypeScript has no ITK, so it mirrors the two nested cases over an interior bump whose reached point is known in closed form. A constant field would not do: its uniform shift is visible from the boundary, so the walk finds the right answer by accident.

Sensitivity checked by neutralizing the accumulation in each port: exactly the nested cases fail, the other five pass either way, being the ones the fold already handled.

Python 1502 passed, TypeScript 719, prek and deno clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resampling bounds for composite transforms that combine multiple displacement fields, B-spline stages, and affine transformations.
  * Preserved accumulated displacement ranges across nested and sequential transform stages.
  * Improved physical-region containment for complex transformed grids, including interior displacement variations.

* **Tests**
  * Added coverage for composite transforms, multiple nonlinear stages, affine compositions, and displacement-field measurements.
  * Added validation for constant shifts and localized displacement changes.

### DIFF
--- a/py/ngff_zarr/resample_bounding_box.py
+++ b/py/ngff_zarr/resample_bounding_box.py
@@ -538,22 +538,39 @@ def _box_inside_field_domain(entry, dimension: int, grid_box) -> bool:
     return True
 
 
-def _folded_interval(interval, outer_entries, dimension: int):
-    """``interval`` carried through the stages applied after its own.
+def _accumulated_interval(entries, intervals, dimension: int, contained: bool):
+    """The range the displacement stages of ``entries`` can add, or ``None``.
 
-    ITK applies the last entry of a list first, so the stages applied after
-    entry ``i`` are the entries before it. An affine stage maps the interval
-    through the signs of its matrix; an orthonormal stage bounds every output
-    component by the Euclidean reach of the input; anything else returns
+    ITK applies the last entry of a list first, so the walk runs from the
+    last entry outward, carrying what the stages seen so far can add. A
+    displacement stage adds its own range, since the linear list this widens
+    replaced that stage by the identity; an affine stage maps the range
+    through the signs of its matrix; an orthonormal stage bounds every
+    component by the range's Euclidean reach. Any other stage returns
     ``None``, and the caller keeps the boundary walk.
+
+    Accumulating outward rather than folding each stage separately is what
+    lets a displacement stage sit outside another one: the range reaching it
+    is carried through, and its own range joins on top.
+
+    :param intervals: One :func:`_stage_interval` per entry, so a field's
+        parameters are scanned once for the whole list.
     """
     from .itk_transform_to_ngff_transform import (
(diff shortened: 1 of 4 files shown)


thewtex (approved):
🙏 💯

PR #736, opened 2026-09-03
Store a shard as asked, and let the caller keep the axis order

### LINKED ISSUE #733: Unexpected chunk and shard sizes
Using latest ngff-zarr:

```
$ pip freeze | grep zarr
ngff-zarr==0.45.0
zarr-metadata==0.5.0
zarrista==0.1.0
```

With this script to read the image at https://ome.github.io/ome-ngff-validator/?source=https://livingobjects.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr 
where the largest resolution has:

```
Shape: 2,236,275,271
Chunk: 1,1,256,256
Shard: 1,10,512,512
```

```
import ngff_zarr as nz

multiscales = nz.from_ngff_zarr('https://livingobjects.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr')

nz.to_ngff_zarr("6001240_shards.ome.zarr", multiscales, chunks_per_shard={"z": 10, "y": 2, "x": 2, "c": 2})
```

The resulting image has treated the `x` and `y` chunk sizes quite differently.
Instead of the chunks being "square", I see

```
Chunk: 1,1,55,271
Shard: 2,10,275,271
```

Maybe this is the expected behaviour (because the image is kinda small?), but I just wanted to check / understand a bit better?

Thanks!

### LINKED ISSUE #734: Testing 0.9.dev1 dimension order
Hi,
 I was trying to create some sample `0.9.dev1` data with a non-standard dimension order (as allowed under RFC-3)...

However, I found that `nz.to_multiscales()` seems to transpose my `Z,Y,X,C` NgffImage back to `C,Z,Y,X`.
I understand that up until now, this would have been very useful behaviour, but I'm wondering if this should be relaxed now, or at least let the user have some control over it?

Here's my script...

```
import ngff_zarr as nz
import numpy as np

from zarrista import Array
from zarrista.store import FilesystemStore

store = FilesystemStore("6001240.ome.zarr")
array = Array.open(store, path="/0")

# convert zarrista Array to numpy array...
zarrista_data = array[array.subset_all]
numpy_arr = zarrista_data.to_numpy()

# Swap dimensions from C,Z,Y,X to Z,Y,X,C
arr2 = np.transpose(numpy_arr, (1, 2, 3, 0))

# image NgffImage(data=dask.array<array, shape=(236, 275, 271, 2),...
image = nz.to_ngff_image(arr2, dims=['z', 'y', 'x', 'c'], scale={'z': 0.3, 'y': 0.3, 'x': 0.3})

# transposed back to C,Z,Y,X here
# NgffMultiscales(images=[NgffImage(data=dask.array<rechunk-merge, shape=(2, 236, 275, 271),.. dims=('c', 'z', 'y', 'x')
multiscales = nz.to_multiscales(image, scale_factors=[2,4])
nz.to_ngff_zarr("6001240_ZYXC.ome.zarr", multiscales, version="0.9.dev1")
```

Many thanks for the nice library and the early support of `0.9.dev1`!

vboussot:
Two regressions reported within a day of each other, both in how the writer reinterprets what it was given.

## A shard is stored as requested (closes #733)

A sharded write shortened every shard dimension to the axis length. The sharding codec requires the inner chunk to divide the shard, so the inner chunk was then pulled down to a divisor of that shortened shard. An axis of 275 (5 * 5 * 11) has no divisor near a requested 256, giving 55, and a prime axis of 271 leaves only the whole shard, giving 271. A store asked for 256 x 256 chunks in 512 x 512 shards came back with 1 x 1 x 55 x 271 chunks in 2 x 10 x 275 x 271 shards.

Zarr v3 permits a partial final shard exactly as it permits a partial final chunk, which is already why chunks are no longer snapped to an axis divisor (#588). The shard is now stored as requested, held only to the smallest whole number of inner chunks that covers the axis, so no shard grid is larger than the array needs and the inner chunk divides it by construction.

The region planner keeps working on the shard clamped to the axes, since a write region cannot run past the array. Conflating the two is what made the large-array path divide by zero once shards could exceed it.

Checked: the three write paths (small, large, metadata-only) describe the array identically, which was the point of the change that introduced this; the data round-trips bit for bit on the small and large paths; zarr-python reads the partial final shard and returns the same array; #588 does not reopen.

## The axis order can be kept (closes #734)

`to_multiscales` reordered any input whose axes were not in the specification order, so a `(z, y, x, c)` image came back as `(c, z, y, x)` with no way to object. RFC-3 lifts the ordering rule from 0.9.dev1 on. The guard meant to leave RFC-3 models alone only fires on an axis *name* outside `{t, c, z, y, x}`, which reordered standard names never are.

`axis_order`, `axisOrder` in the TypeScript port, chooses:

- `"canonical"`, the default and the current behaviour, reorders to time, then channel, then space. Conversion sources need it: the TIFF `S` axis, ITK component images and the 4-D/5-D default dims inference all yield channel-last.
- `"preserve"` writes the axes as given.

The writer still refuses a non-canonical order for every version before 0.9.dev1, with the message it already had, so `"preserve"` cannot produce a store an earlier version would reject.

## Note

`to_ngff_image` rejects any axis name outside `{t, c, z, y, x}`, so an RFC-3 axis model with its own names is only constructible through `NgffImage` directly, with `axes_types`. That is a separate gap and is left alone here.

## Summary by CodeRabbit

- **New Features**
  - Added an option to preserve input axis order when converting data to multiscale OME-Zarr. Canonical OME-Zarr ordering remains the default.
  - Added validation for unsupported axis-order settings.

- **Bug Fixes**
  - Improved sharded storage resolution so requested shard dimensions are preserved, including when they extend beyond array boundaries.
  - Region planning now correctly caps chunk coverage at the relevant array dimensions.

- **Tests**
  - Added coverage for axis-order preservation, validation, round-tripping, and sharding edge cases.

### DIFF
--- a/py/ngff_zarr/to_multiscales.py
+++ b/py/ngff_zarr/to_multiscales.py
@@ -10,7 +10,7 @@ import uuid
 from collections.abc import Mapping, MutableMapping, Sequence
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any
+from typing import Any, Literal
 
 import numpy as np
 from dask.array.core import Array as DaskArray
@@ -438,6 +438,7 @@ def to_multiscales(
     progress: NgffProgress | NgffProgressCallback | None = None,
     cache: bool | None = None,
     orientation: str | Mapping[str, AnatomicalOrientation] | None = None,
+    axis_order: Literal["canonical", "preserve"] = "canonical",
 ) -> NgffMultiscales:
     """
     Generate multiple resolution scales for the OME-NGFF standard data model.
@@ -471,10 +472,17 @@ def to_multiscales(
         output automatically whenever it is present; no separate opt-in is required.
     :type  orientation: str or mapping of str to AnatomicalOrientation, optional
 
+    :param axis_order: What to do with an input whose axes are not in the
+        OME-Zarr specification order. ``"canonical"``, the default, reorders
+        them to time, then channel, then space (t, c, z, y, x) with a lazy
+        transpose, which is what conversion sources need: the TIFF ``S``
+        (sample) axis, ITK component images and the 4-D/5-D default dims
+        inference all yield channel-last. ``"preserve"`` writes the axes in
+        the order given, which RFC-3 permits from OME-Zarr 0.9.dev1 on and
+        which no earlier version accepts.
+    :type  axis_order: str, optional
+
     :return: NgffImage for each resolution and NGFF multiscales metadata.
-        Axes are normalized to the OME-Zarr specification order -- time, then
-        channel, then space (t, c, z, y, x) -- with a lazy transpose when the
-        input image orders them differently.
     :rtype : NgffMultiscales
     """
     # Shallow copy, with its own computed_callbacks: the rechunk and dask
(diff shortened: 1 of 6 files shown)


thewtex (approved):
@vboussot thank you!

PR #737, opened 2026-09-03
Don't force metadata consolidation

### LINKED ISSUE #698: Allow **not** consolidating metadata
# Problem

`to_ngff_zarr` currently unconditionally consolidates metadata:

https://github.com/fideus-labs/ngff-zarr/blob/62bdd307eabd02d62e825ca949d9a1bf56e3d029/py/ngff_zarr/to_ngff_zarr.py#L1336

which makes it incompatible with stores that do not allow this. For example Icechunk does not allow consolidated metadata as it interferes with Icechunk's own consolidation and consistency mechanisms (see discussion here: https://github.com/zarr-developers/zarr-python/pull/3119)

https://github.com/earth-mover/icechunk/blob/224519be49795c91fa35c2404b04ad5cb298a065/icechunk/src/store.rs#L265-L266

# Solution

1. Automatic

only consolidate if the store advertises that it supports it.

```python
if getattr(store, "supports_consolidated_metadata", True):
```

2. User choice (as an additional lever on top of 1)

Allow the user to chose this behavior via a kwarg

ianhi:
Don't force metadata consolidation. Fixes #698 

There's a little bit more complexity than I thought at first. We have to protect the user against accidentally creating a store with a stale consolidated metadata when they append. If they wrote a store with consolidated metadata, then append without consolidated metadata they could end up with stale metadata.

This is only a problem for zarr v2 as zarrista re-writes the entire zarr.json document for zarr v3. One of the tests added accounts for the dependence on that implementation detail. But happy to also throwing an error in this case if you prefer that.

## Summary by CodeRabbit

* **New Features**
  * Added an option to control whether metadata is consolidated when converting or optimizing OME-Zarr stores.
  * Metadata consolidation remains enabled by default.
  * Stores can now be written without consolidated metadata for compatibility with backends that reject it.
  * In-place Zarr v2 writes now report an error when disabling consolidation could leave stale metadata.

### DIFF
--- a/mcp/ngff_zarr_mcp/models.py
+++ b/mcp/ngff_zarr_mcp/models.py
@@ -82,6 +82,13 @@ class ConversionOptions(BaseModel):
     use_tensorstore: bool = Field(
         False, description="Deprecated: uses the zarrista backend for I/O"
     )
+    consolidate_metadata: bool = Field(
+        True,
+        description=(
+            "Write consolidated metadata. Set False for stores destined for a "
+            "backend that rejects it, such as Icechunk"
+        ),
+    )
 
     # Performance options
     use_local_cluster: bool = Field(
@@ -162,6 +169,13 @@ class OptimizationOptions(BaseModel):
     )
     storage_options: dict[str, str | int | bool] | None = Field(
         None, description="Storage options for remote stores"
+    )
+    consolidate_metadata: bool = Field(
+        True,
+        description=(
+            "Write consolidated metadata. Set False for stores destined for a "
+            "backend that rejects it, such as Icechunk"
+        ),
     )
 
     @model_validator(mode="after")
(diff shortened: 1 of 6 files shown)


mcp/ngff_zarr_mcp/models.py
```diff
@@ -82,6 +82,13 @@ class ConversionOptions(BaseModel):
     use_tensorstore: bool = Field(
         False, description="Deprecated: uses the zarrista backend for I/O"
     )
+    consolidate_metadata: bool = Field(
+        True,
+        description=(
+            "Write consolidated metadata. Set False for stores destined for a "
+            "backend that rejects it, such as Icechunk"
```
ianhi, line 89: Called out icechunk explicitly here as it's the only one I'm immediately aware of that this applies to. But looking back on it maybe we should remvoe that?

thewtex: @ianhi thanks for the contribution!

Icechunk call-out is welcome :-)

thewtex (approved):

PR #741, opened 2026-09-08
fix(py): carry RFC-4 orientation onto the image on read

vboussot:
`from_ome_zarr` left `NgffImage.axes_orientations` unset and only kept the raw axis entry on `metadata.axes`, so `ngff_image_to_itk_image` built an identity direction for a store that declares its orientation. The TypeScript reader already fills `axesOrientations`.

- `rfc4.axes_orientations_from_axes` derives the orientations from the parsed axes; only an anatomical orientation with a value from the RFC 4 vocabulary is carried, the raw entry stays on `metadata.axes` for validation.
- The v0.4 and v0.6 readers pass it to every `NgffImage`.
- Tests: LPS and RAS read back at 0.4 and 0.5, the absent case, and a malformed orientation read without validation.

Full Python suite: 1549 passed, 5 skipped.

## Summary by CodeRabbit

- **New Features**
  - Added support for reading RFC 4 anatomical axis orientations from OME-Zarr 0.4 and 0.5 metadata.
  - Propagated recognized axis orientations to associated images during metadata loading.
  - Supports orientation data provided in standard axis metadata formats.

- **Bug Fixes**
  - Invalid, unsupported, or incomplete orientation entries are safely ignored during non-validating reads.
  - Images remain without orientation data when no valid orientations are present.

- **Tests**
  - Added integration coverage for orientation round-tripping and malformed metadata handling.

### DIFF
--- a/py/ngff_zarr/rfc4.py
+++ b/py/ngff_zarr/rfc4.py
@@ -8,6 +8,7 @@ to OME-NGFF axes, based on the LinkML schema.
 
 from __future__ import annotations
 
+from collections.abc import Sequence
 from dataclasses import dataclass
 from enum import StrEnum
 from typing import Literal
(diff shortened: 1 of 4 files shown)


thewtex (approved):
👍

